PR #12, opened 2020-11-22
GeekBench support

vosen:
This is the tracking issue for failing GeekBench workloads:

- [x] Depth of Field\
Seems to be a problem with `TRIVIAL_RA`. Works with `IGC_VISAOptions="-nolocalra"` environment variable which effectively disables `TRIVIAL_RA`. Need to report to IGC team
- [x] Histogram Equalization
- [x] Face Detection
- [x] Feature Matching
- [x] Particle Physics
- [x] SFFT\
All five above currently give very worrying  debug spew (if the compiler was built in `DEBUG`) about assembly operand regions being wider than 2 GRFs. Currently pending more debug. Additionally Face Detection and Particle Physics hang with `IGC_VISAOptions="-nolocalra" `.

### DIFF
--- a/zluda/src/impl/function.rs
+++ b/zluda/src/impl/function.rs
@@ -26,6 +26,7 @@ pub struct FunctionData {
     pub arg_size: Vec<usize>,
     pub use_shared_mem: bool,
     pub properties: Option<Box<l0::sys::ze_kernel_properties_t>>,
+    pub do_nothing_hack: bool,
 }
 
 impl FunctionData {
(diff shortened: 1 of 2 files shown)


vosen:
Problem with `TRIVIAL_RA` has been reported internally. Other kernels are now no longer emitting any error message, so the problem might be with ZLUDA, will debug.

vosen:
Problem with Depth of Field is actually a bug in ZLUDA. After #20 lands there's going to be a PR with a fix.
The issue is basically that we don't do conversion between 32bit PTX ntid/ctid etc. and equivalent 64bit SPIR-V builtins. I thought compiler does silent conversion, but instead it does silent corruption!

vosen:
Now, with #20 and #21 everything except Feature Matching works out of the box on master. For some reason Particle Physics and SFFT have really poor performance.
More debugging in progress

vosen:
Closing. Once #36 is merged, Geekbench will fully work

PR #18, opened 2020-11-30
Add a library for dumping kernels arguments before and after launch

vosen:
Currently, I debug NVIDIA kernels by a combination of manual dumping of kernel arguments from within GDB and using opencl-intercept-layer on GeekBench OpenCL. This is a massive pain. Since there's no equivalent of opencl-intercept-layer for CUDA, I'll do a limited version myself.
This is purely for debugging and not finished yet.


vosen:
One feature left: dumping module source alongside kernel enqueues

PR #25, opened 2020-12-25
Add building only CI

take-cheeze:
Testing isn't working yet because cuda stuff needs to be installed.
Succeeded log: https://github.com/take-cheeze/ZLUDA/runs/1607094707


vosen:
Neat, thanks.
For the record, to have tests running we will need a machine that has NVIDIA GPU, because we verify some host-side behavior against a live CUDA installation with a live GPU (edge case behavior on the host side is relatively non-intuitive and under-documented). A machine that has *both* Intel GPU and NVIDIA GPU would be the best.
I hope to procure such machine for CI some time soon.

PR #34, opened 2021-01-20
Add script for replaying dumped kernel

vosen:
In #18 we gained ability to dump before/after arguments of CUDA kernel launches. Now, this script will add ability to actually execute this dump and verify if results are correct


vosen:
Missing features:
* Accepting path from argument
* Handling both single-launch dir dump and whole-program-launch dump dir

PR #61, opened 2024-02-12
Tidy up some English in ARCHITECTURE.md

smcl:
In the paragraph on the NVidia Dark API there were a handful of minor wording issues - missing or duplicated articles (like "a" and "the"), and use of "tens of X" when "dozens of X" is more idiomatic. Very minor stuff - it's readable without my changes.

If this kind of contribution is not welcome, then I apologise. If it is welcome, then I can happily take a closer look at the rest of the text - I didn't want to just dump a massive PR unannounced.


vosen:
Thanks. Fixes of any kind are always welcome.
For my defense, I frankly thought nobody would willingly read it

PR #196, opened 2024-04-05
Implement mad.hi.cc

NyanCatTW1:
Ran into a mad.hi.cc.u32 while trying to get a certain HPC software working. This PR implements the opcode similarly to MadC handling as well as adding a simple test for mad.hi.cc
The aforementioned software is still producing incorrect result despite this patch. I'm working on finding other underlying causes and will open subsequent PRs when I spot other issues

### DIFF
--- a/ptx/src/emit.rs
+++ b/ptx/src/emit.rs
@@ -2099,21 +2099,6 @@ fn emit_inst_madcc(
         mul_result,
         arg.src3,
     )
-    /*
-        let builder = ctx.builder.get();
-        let src1 = ctx.names.value(arg.src1)?;
-        let src2 = ctx.names.value(arg.src2)?;
-        let mul_result = unsafe { LLVMBuildMul(builder, src1, src2, LLVM_UNNAMED) };
-        emit_inst_addsub_cc_impl(
-            ctx,
-            "add",
-            type_,
-            arg.dst,
-            arg.carry_out,
-            mul_result,
-            arg.src3,
-        )
-    */
 }
 
 fn get_llvm_type_struct<'a>(
@@ -2192,29 +2177,6 @@ fn emit_inst_madc(
         mul_result,
         args.src3,
     )
-    /*
-       let src3 = ctx.names.value(args.src3)?;
-       let add_no_carry = unsafe { LLVMBuildAdd(builder, mul_result, src3, LLVM_UNNAMED) };
-       let carry_flag = ctx.names.value(args.carry_in)?;
-       let llvm_type = get_llvm_type(ctx, &ast::Type::Scalar(type_))?;
-       let carry_flag = unsafe { LLVMBuildZExt(builder, carry_flag, llvm_type, LLVM_UNNAMED) };
-       if let Some(carry_out) = args.carry_out {
-           emit_inst_addsub_cc_impl(
-               ctx,
-               "add",
-               type_,
-               args.dst,
-               carry_out,
-               add_no_carry,
-               carry_flag,
-           )?;
-       } else {
-           ctx.names.register_result(args.dst, |dst| unsafe {
-               LLVMBuildAdd(builder, add_no_carry, carry_flag, dst)
-           });
-       }
-       Ok(())
-    */
 }
 
 fn emit_inst_add_c(
--- a/ptx/src/emit.rs
+++ b/ptx/src/emit.rs
@@ -2079,17 +2079,17 @@ fn emit_inst_mad_lo(
     )
 }
 
-// TODO: support mad.hi.cc
 fn emit_inst_madcc(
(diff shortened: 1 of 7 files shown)


ptx/src/test/spirv_run/mad_hi_cc.ptx
vosen, line 1: Please simplify this test to just check if `mad.hi` sets carry flag correctly, so just:
````
mad.hi.cc.u32 ... // non-overflowing
addc.u32            carry_out_1, 0, 0; // carry_out_1 should be 0
mad.hi.cc.s32 ... // overflowing
addc.u32            carry_out_2, 0, 0; // carry_out_2 should be 1
````
without all the suft from `// set carry=1` and below

👍 It's good that you want to test extensively, but since those complex scenario are already present in `madc_cc2` and `subc_cc2` let's not duplicate them

File: ptx/src/emit.rs [hunk 1]
```diff
@@ -2098,6 +2099,21 @@ fn emit_inst_madcc(
         mul_result,
         arg.src3,
     )
+    /*
```
vosen: Remove this comment. Feel free to also remove the comment in `emit_inst_madc`, I forgot to remove it earlier

vosen (changes requested):

vosen:
Thanks, good work. Please make some small changes.

I've looked at all madc/subc/*.cc code and I'm starting to think that subc/sub.cc code is wrong. Maybe that's the reason? I'll double check it after your PR merges

NyanCatTW1:
> Thanks, good work. PR requires just some small improvements.
> 
> I've looked at all madc/subc/*.cc code and I'm starting to think that subc/sub.cc code is wrong. Maybe that's the reason? I'll double check it after your PR merges

It seems like CUDA somehow miscompiles CC calls often. Here is a minimal repro that I coincidentally just finished a few minutes ago:
```
// ...
// Trigger CUDA miscompile
mov.u32         const1, 1;
mov.u32         zero, 0;
add.cc.u32      unused, zero, const1;

// Perform a trivial computation that should've yielded 0
subc.cc.u32 result_1, zero, zero;

// Trigger CUDA miscompile
mov.u32         const1, 1;
mov.u32         zero, 0;
add.cc.u32      unused, zero, const1;

// Perform a typical computation
subc.cc.u32 result_2, zero, src1;
// ...
```

ZLUDA: `result_1: 0x0, result_2: 0xB29A9089`
CUDA 12.4: `result_1: 0xFFFFFFFF, result_2: 0xB29A9088`

Full repro: [cc_cf_quirks.tar.gz](https://github.com/vosen/ZLUDA/files/14885065/cc_cf_quirks.tar.gz)
Compile with `nvcc -lcuda -lcudart cc_cf_quirks.cu -o cc_cf_quirks`

vosen:
Yep, that's what I'm talking about. I don't think it's a miscompilation.
Current ZLUDA interpretation is that there are two flags: underflow and overflow, but it might be that subc subtracts 1 if carry flag is 0 and subtracts 0 if carry flag is 1. As I said, I'll look into it more closely

NyanCatTW1:
> Yep, that's what I'm talking about. I don't think it's a miscompilation. Current ZLUDA interpretation is that there are two flags: underflow and overflow, but it might be that subc subtracts 1 if carry flag is 0 and subtracts 0 if carry flag is 1. As I said, I'll look into it more closely

Here is another data point in case it turns out to be helpful:
In https://github.com/NVlabs/CGBN/blob/e8b9d265c7b84077d02340b0986f3c91b2eb02fb/include/cgbn/core/padded_resolver.cu#L69, changing `chain.add(c, 0xFFFFFFFF);` into `chain.sub(c, 1);` retains the same CUDA behavior but "fixes" the ZLUDA behavior. I have no idea how that works, but I know that chain.add uses add.cc.u32, whereas chain.sub uses sub.cc.u32

vosen:
subc/sub.cc behavior should be fixed as of #197

vosen:
And with #198 most of CGBN tests pass. Everything division related fails, this is an incomplete gtest filter I used to filter out failing tests:
```
--gtest_filter=-S288T8/CGBN1/0.div_1:S288T8/CGBN1/0.rem_1:S288T8/CGBN1/0.div_rem_1:S288T8/CGBN3/0.div_wide_1:S288T8/CGBN3/0.rem_wide_1:S288T8/CGBN3/0.div_rem_wide_1:S288T8/CGBN5/0.modular_power_1:S288T8/CGBN5/0.bn2mont_1:S288T8/CGBN5/0.barrett_div_1:S288T8/CGBN5/0.barrett_rem_1
```
Feel free to look into CGBN now, I won't be debugging it in the next few days

PR #211, opened 2024-04-23
Fix + improve vprintf implementation

NyanCatTW1:
We were trying to debug a kernel with printf, but then we realized %d and friends didn't work at all! So we dug in and made changes to vprintf that should allow it to support all sorts of common printf calls

Test code (mostly taken from https://cplusplus.com/reference/cstdio/printf/)
```c
#include <stdio.h>
#include <cuda.h>

__global__ void kernel() {
    printf("Characters: %c %c \n", 'a', 65);
    printf("Decimals: %d %ld\n", 1977, 650000L);
    printf("Preceding with blanks: %10d \n", 1977);
    printf("Preceding with zeros: %010d \n", 1977);
    printf("Some different radices: %d %x %o %#x %#o \n", 100, 100, 100, 100, 100);
    printf("floats: %4.2f %+.0e %E \n", 3.1416, 3.1416, 3.1416);
    printf("Width trick: %*d \n", 5, 10);
    printf("%s \n", "A string");
    printf("I'm not formatting anything: %%s!\n");
}
```

Expected output (currently identical to CUDA's)
```
Characters: a A 
Decimals: 1977 650000
Preceding with blanks:       1977 
Preceding with zeros: 0000001977 
Some different radices: 100 64 144 0x64 0144 
floats: 3.14 +3e+00 3.141600E+00 
Width trick:    10 
A string 
I'm not formatting anything: %s!
```

Some of the code could've been done better, but well, at least now it works properly


ptx/lib/zluda_ptx_impl.cpp
```diff
@@ -1393,8 +1395,36 @@ extern "C"
...
+
+            // %% requires no additional handling
+            if (*s == '%')
+            {
+                s++;
+                continue;
+            }
+
+            // %s uses __ockl_printf_append_string_n
+            // https://github.com/ROCm/clr/blob/5a715ed160c69a2deebac187a0368900a774f3e8/hipamd/include/hip/amd_detail/hip_assert.h#L73
```
vosen: It's going to be more informative to instead link to ockl sources: https://github.com/ROCm/ROCm-Device-Libs/blob/rocm-5.7.x/ockl/src/services.cl#L343

ptx/lib/zluda_ptx_impl.cpp
```diff
@@ -1406,17 +1436,21 @@ extern "C"
...
                     {
                         s += specifier_with_length;
-                    }
-                    if (len > 0)
-                    {
-                        uint64_t value = read_valist(valist_ptr, valist_offset, len);
-                        handle = __ockl_printf_append_args(handle, 1, value, 0, 0, 0, 0, 0, 0, 0);
+                    } else {
+                        s++;
+                        continue;
```
vosen, line 1442: This is the case of unkown/invalid specifier, e.g. `%!`. Shouldn't this be a `break` instead?

ptx/lib/zluda_ptx_impl.cpp
```diff
@@ -1406,17 +1436,21 @@ extern "C"
...
+                if (len > 0)
+                {
+                    uint64_t value = read_valist(valist_ptr, valist_offset, len);
+                    handle = __ockl_printf_append_args(handle, 1, value, 0, 0, 0, 0, 0, 0, 0);
+                }
+                break;
             }
         }
         return (uint32_t)__ockl_printf_append_args(handle, 0, 0, 0, 0, 0, 0, 0, 0, 1);
-        return 1;
```
vosen, line 1419: Please change so it does not return `__ockl_printf_append_args`, but 1 instead. `__ockl_printf_append_args` returns an arbitrary handle. I don't know what I were thinking

vosen (changes requested):
Mostly good, but I need some small changes

vosen:
> Some of the code could've been done better, but well, at least now it works properly


Yes, te vprintf code is not very good. And difficult to test properly: last time I tried to capture from stdout handle it didn't work with HIP.
I should probably split printf into a separate file and test it on CPU with mocked `__ockl_printf_*`, but it's a story for another day.

ptx/lib/zluda_ptx_impl.cpp
NyanCatTW1: Can't do that, as it'll make the current oversimplified handling of sub-specifier (aka. ignoring them and assuming they are valid) break:
```
Characters: a A 
Decimals: 1977 650000
Preceding with blanks: Preceding with zeros: Some different radices: 100 64 144 floats: 3.14 Width trick:    10 
A string
```

While using continue; here might break unknown/invalid specifier handling, considering we've implemented the entirety of the standard specifiers + handling of invalid specifiers aren't strictly defined, I find it a good compromise

vosen: Ah, good point

vosen (approved):

PR #227, opened 2024-05-08
Updated Rust version to 1.70.0

Michael-Hollister:
When building the docker container I got the following error:
`error: package lalrpop v0.20.2 cannot be built because it requires rustc 1.70 or newer, while the currently active rustc version is 1.66.1`

Updated Rust version in the dockerfile to fix devcontainer build.


vosen:
Hi, I'm closing it because I opted for a different approach in #229 that preserves compatibility with rust 1.66.1.
If, despite this, build problems persist please open another issue or PR

Michael-Hollister:
No problem, that works as well and had no issues building with the new changes. Thanks!

PR #237, opened 2024-05-23
Update README with CGBN information

vosen:


vosen:
Pinging @NyanCatTW1 who was interested in CGBN

PR #267, opened 2024-08-14
PTX parser rewrite

vosen:
Our PTX parser has been working surprisingly well and has been the same basically from the inception.
Lately, I've been running into limitations of the parser when working on other areas of the codebase. This rewrite is done for those explicit reasons (not all of them will apply right now):
* Better support of inline assembly (out-of-order instruction modifiers)
* More precise tracking of flush-to-zero flag
* Better support for instruction visitors (we can only do move-and-map right now, but we want visit-by-ref and visit-by-mut-ref)
* Less error-prone instruction definitions (no more `ArgumentDescriptor`)

Side benefits:
* Slightly better compilation times
* PTX parser will be split to a separate crate, making it easier for people who want to use just that part of ZLUDA


erkinalp:
v4-reborn 🚀

PR #277, opened 2024-09-25
Re-enable all failing PTX tests

vosen:


vosen:
Certain parts are a bit dodgy (`cvt`), but they will be refined later with better tests

Issue #285, opened 2024-10-13
Improve host code bindings


vosen:
Expected outcome:
Current host code APIs in `zluda` does not use generated bindings: every CUDA update requires manually reapplying the functions we expose. Instead `zluda` should use `cuda_function_declarations` macro, same as `zluda_dump`.
While at it, `CUresult` should be redefined to effectively be `Result<(), NonZeroU32>`

Comments:
* Probably should be done together with #286

PR #299, opened 2024-11-19
Fix host code and update to CUDA 12.4

### LINKED ISSUE #286: Update to CUDA 12.4
Expected outcome:
`zluda` and `zluda_dump` use/expose CUDA 12.4 APIs

Comments:
* You need to regenerate in `cuda_base/src/cuda.rs` and fix all the failures
* Probably should be done together with #285

vosen:
Will close #286 and #285


cuda_base/src/cuda.rs
Rayyan98, line 1: bruh

PR #302, opened 2024-11-27
README.md: improve build instructions

thomas-mc-work:
I bet it's not required to have more than a shallow copy (history depth bigger than 1).

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Not supported
 ## Building
 **Warning**: ZLUDA is under heavy development (see news [here](https://vosen.github.io/ZLUDA/blog/zludas-third-life/)). Instructions below might not work.
 
-_Note_: This repo has submodules. Make sure to recurse submodules when cloning this repo, e.g.: `git clone --recursive https://github.com/vosen/ZLUDA.git`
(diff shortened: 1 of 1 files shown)


vosen:
Are you certain that it's required in the command line?
I though setting it in the `.gitmodules` is enough: https://github.com/vosen/ZLUDA/blob/870fed4bb69d919a10822032d65ec20f385df9d7/.gitmodules#L12

thomas-mc-work:
I think this only responsible for the external llvm module. And even this doesn't seem to work as I've just tested to verify. I guess this config can't be evaluated when cloning the repo because git won't read it until it's there.

thomas-mc-work:
WDYT?

vosen:
I've double checked, current situation with `shallow = true` makes so that only submodule gets a shallow checkout, but that's fine. The only submodule we use is LLVM and the way we use LLVM means that we are going to stay on a particular version for years, so no problem with updates. Also LLVM checkout takes by far the most time to checkout.
On the other hand ZLUDA  codebase moves quickly and there's a) no need to build it manually to try it since we have automatic builds now b) if you are a developer and want to build from sources, I'd rather give you by default instructions that are suitable for development
For those reasons I'll be closing this PR

PR #304, opened 2024-12-05
Enable Geekbench 5

vosen:
First working application!


vosen:
For posterity: canny sub-benchmark performance is probably due to ZLUDA being unable to coalesce global loads.
OpenCL compiles down to:
```
global_load_u8 v1, v[1:2], off offset:-1
global_load_u16 v2, v[5:6], off         
global_load_u8 v5, v[3:4], off offset:1 
global_load_u8 v3, v[3:4], off offset:-1
```
ZLUDA:
```
global_load_u8 v10, v10, s[0:1]
global_load_u8 v4, v4, s[0:1]  
global_load_u8 v11, v11, s[0:1]
global_load_u8 v12, v12, s[0:1]
```
This optimization could be exposed behind an unsafe flag, but we would need to check with LLVM what exact pattern we need to emit

PR #315, opened 2024-12-20
Update README

vosen:


README.md
```diff
@@ -24,33 +25,54 @@ Run your application like this:
...
-cargo build --release
-```
-in the main directory of the project.  
 ### Linux
 
-If you are building on Linux you must also symlink (or rename) the ZLUDA output binaries after ZLUDA build finishes:
+If you are building on Linux you must also symlink the ZLUDA output binaries after ZLUDA build finishes:
 ```
+cd target/release
 ln -s libnvcuda.so target/release/libcuda.so
```
JoelleJS: `target/release/` probably needs to be removed from this line and the next two right?

vosen: true, updated now

PR #316, opened 2025-01-01
Fix test zluda_dump

JoelleJS:
I assumed that the new behavior (outputs `CUDA_SUCCESS` instead of `0x0`) is correct and changed the test expectation. Happy to be corrected if wrong.


vosen:
Thanks, you assumed correctly

Issue #312, opened 2024-12-13
Verify LLVM output in ptx unit tests


vosen:
Expected outcome:
For each .ptx file that is unit-tested by the `ptx` project there is a textual LLVM output that is also verified (as a separate Rust unit tests)

Comments:
* Last time I checked LLVM does not ship ability to compare Module objects (and implementing it ourselves would be complex due to all the details). It's sufficient to just compare strings

JoelleJS:
I'd like to have a go at this

vosen:
Closed by #324

PR #324, opened 2025-01-15
LLVM unit tests

JoelleJS:
#312


JoelleJS:
@vosen Yep, still working on that part. Just wanted the .ll files committed

JoelleJS:
Okay, the tests seem to be working individually, e.g. `cargo test -p ptx -- ::add_llvm`. However, when running `cargo test -p ptx -- llvm --nocapture` I can see that it only runs a few tests and then fails with SIGSEGV or SIGABRT.

I'm thinking there is some kind of memory leak, but I haven't been able to find it. Perhaps your trained eyes see it quicker @vosen

vosen:
I'm not at a computer right now, but I suspect LLVMParseBitcode2. Try
LLVMParseBitcodeInContext2.
LLVM C API has a bunch of those function which don't take an explicit
context parameter and use some sort of  default global context. That
explodes when mixed with code that uses explicit context (like ZLUDA).
It's a horrible default and I was bitten by it previously

śr., 5 lut 2025, 16:32 użytkownik Joëlle van Essen ***@***.***>
napisał:

JoelleJS:
@vosen That was exactly it, good catch.

All tests succeed, as expected (the `.ll` files were generated from the LLVM IR produced in the tests). The assumption is that the files are currently correct.

Currently, when the assertion fails it looks like this:
```
assertion `left == right` failed
  left: "declare i32 @__zluda_ptx_impl_sreg_tid(i8) #0\n\ndeclare i32 @__zluda_ptx_impl_sreg_ntid(i8) #0\n\ndeclare i32 @__zluda_ptx_impl_sreg_ctaid(i8) #0\n\ndeclare i32 @__zluda_ptx_impl_sreg_nctaid(i8) #0\n\ndeclare i32 @__zluda_ptx_impl_sreg_clock() #0\n\ndeclare i32 @__zluda_ptx_impl_sreg_lanemask_lt() #0\n\ndefine amdgpu_kernel void @neg(ptr addrspace(4) byref(i64) %\"34\", ptr addrspace(4) byref(i64) %\"35\") #0 {\n  %\"36\" = alloca i64, align 8, addrspace(5)\n  %\"37\" = alloca i64, align 8, addrspace(5)\n  %\"38\" = alloca i32, align 4, addrspace(5)\n  br label %1\n\n1:                                                ; preds = %0\n  %\"39\" = load i64, ptr addrspace(4) %\"34\", align 4\n  store i64 %\"39\", ptr addrspace(5) %\"36\", align 4\n  %\"40\" = load i64, ptr addrspace(4) %\"35\", align 4\n  store i64 %\"40\", ptr addrspace(5) %\"37\", align 4\n  %\"42\" = load i64, ptr addrspace(5) %\"36\", align 4\n  %\"47\" = inttoptr i64 %\"42\" to ptr\n  %\"41\" = load i32, ptr %\"47\", align 4\n  store i32 %\"41\", ptr addrspace(5) %\"38\", align 4\n  %\"44\" = load i32, ptr addrspace(5) %\"38\", align 4\n  %\"43\" = sub i32 0, %\"44\"\n  store i32 %\"43\", ptr addrspace(5) %\"38\", align 4\n  %\"45\" = load i64, ptr addrspace(5) %\"37\", align 4\n  %\"46\" = load i32, ptr addrspace(5) %\"38\", align 4\n  %\"48\" = inttoptr i64 %\"45\" to ptr\n  store i32 %\"46\", ptr %\"48\", align 4\n  ret void\n}\n\nattributes #0 = { \"amdgpu-unsafe-fp-atomics\"=\"true\" \"no-trapping-math\"=\"true\" \"uniform-work-group-size\"=\"true\" }"
 right: "declare i32 @__zluda_ptx_impl_sreg_tid(i8) #0\n\ndeclare i32 @__zluda_ptx_impl_sreg_ntid(i8) #0\n\ndeclare i32 @__zluda_ptx_impl_sreg_ctaid(i8) #0\n\ndeclare i32 @__zluda_ptx_impl_sreg_nctaid(i8) #\n\ndeclare i32 @__zluda_ptx_impl_sreg_clock() #0\n\ndeclare i32 @__zluda_ptx_impl_sreg_lanemask_lt() #0\n\ndefine amdgpu_kernel void @neg(ptr addrspace(4) byref(i64) %\"34\", ptr addrspace(4) byref(i64) %\"35\") #0 {\n  %\"36\" = alloca i64, align 8, addrspace(5)\n  %\"37\" = alloca i64, align 8, addrspace(5)\n  %\"38\" = alloca i32, align 4, addrspace(5)\n  br label %1\n\n1:                                                ; preds = %0\n  %\"39\" = load i64, ptr addrspace(4) %\"34\", align 4\n  store i64 %\"39\", ptr addrspace(5) %\"36\", align 4\n  %\"40\" = load i64, ptr addrspace(4) %\"35\", align 4\n  store i64 %\"40\", ptr addrspace(5) %\"37\", align 4\n  %\"42\" = load i64, ptr addrspace(5) %\"36\", align 4\n  %\"47\" = inttoptr i64 %\"42\" to ptr\n  %\"41\" = load i32, ptr %\"47\", align 4\n  store i32 %\"41\", ptr addrspace(5) %\"38\", align 4\n  %\"44\" = load i32, ptr addrspace(5) %\"38\", align 4\n  %\"43\" = sub i32 0, %\"44\"\n  store i32 %\"43\", ptr addrspace(5) %\"38\", align 4\n  %\"45\" = load i64, ptr addrspace(5) %\"37\", align 4\n  %\"46\" = load i32, ptr addrspace(5) %\"38\", align 4\n  %\"48\" = inttoptr i64 %\"45\" to ptr\n  store i32 %\"46\", ptr %\"48\", align 4\n  ret void\n}\n\nattributes #0 = { \"amdgpu-unsafe-fp-atomics\"=\"true\" \"no-trapping-math\"=\"true\" \"uniform-work-group-size\"=\"true\" }"
```

These can be copied and `diff`ed of course, but I think it would be more convenient to use the `pretty_assertions` crate ([link](https://docs.rs/pretty_assertions/latest/pretty_assertions/)) to show a clear diff in the output itself. Do you agree?

JoelleJS:
Marked this as ready but I just realized I forgot the following point:
> If there's the a variable set (pick a name you like) that points to a directory then it also is written to the directory.

Will do that soon.

JoelleJS:
Done as far as I can see, ready for review 🙂

ptx/src/pass/emit_llvm.rs
```diff
@@ -148,6 +150,21 @@ impl std::fmt::Debug for Message {
 
 pub struct MemoryBuffer(LLVMMemoryBufferRef);
 
+impl MemoryBuffer {
+    pub fn print_as_asm(&self) -> &str {
```
vosen: There's a better approach: instead of doing `LLVMParseBitcodeInContext2` it'd be much easier to change `pass::to_llvm_module`. Currently it returns an object that wraps already compiled and converrted to bitcode LLVM memory buffer. You should instead change that function to return an object that contains both LLVM Context and LLVM Module. All the existing consumers can use `write_bitcode_to_memory` and this one can use a new `print_module_to_string` function.
Furthermore, this code leaks leaking text string: wrap return value from `LLVMPrintModuleToString` in an Message object (defined above) so it does dispose correctly (and it already supports printing with Debug trait, feel free to change to/add Display or add &str getter if required).

ptx/src/test/spirv_run/mod.rs
```diff
@@ -224,6 +238,34 @@ fn test_hip_assert<
...
+    'a,
+>(
+    name: &str,
+    ptx_text: &'a str,
+    expected_ll: &str
+) -> Result<(), Box<dyn error::Error + 'a>> {
+    let ast = ptx_parser::parse_module_checked(ptx_text).unwrap();
+    let llvm_ir = pass::to_llvm_module(ast).unwrap();
+    let actual_ll = llvm_ir.llvm_ir.print_as_asm();
+    let result = catch_unwind(||
```
vosen: Hmmm, that catch_unwind is really unfortunate. I can see that crate `pretty_assertions` also exports `StrComparison` type which seem to wrap the presentation of the diff.
Please check if we can get the same effect and avoid touching unwinding with `if actual_ll != expected_ll { panic!("{}", pretty_assertions::StrComparison::new(actual_ll, expected_ll)); }`

ptx/src/test/spirv_run/mod.rs
```diff
@@ -224,6 +238,34 @@ fn test_hip_assert<
...
+        pretty_assertions::assert_eq!(actual_ll, expected_ll));
+    if let Err(cause) = result {
+        // Write actual generated LLVM IR to directory specified by environment variable
+        // TEST_PTX_LLVM_FAIL_DIR if test fails
+        let output_dir = env::var("TEST_PTX_LLVM_FAIL_DIR");
+        if let Ok(output_dir) = output_dir {
+            let output_dir = Path::new(&output_dir);
+            create_dir_all(&output_dir).unwrap();
```
vosen: (Optional) My preference is that free functions come with the namespace, so `fs::create_dir_all` instead of `create_dir_all`

vosen (changes requested):

JoelleJS:
Hi @vosen, I have made changes that hopefully address your comments.

JoelleJS:
Upon further thought, returning a tuple from `emit_llvm::run` leaves the responsibility of making sure that the context survives the lifetime of the module to the caller. I think that's what you were implying in your comment. So I should still wrap that in an object. Do you think it makes sense to include the context in the `Module` object that now only wraps an `LLVMModuleRef`, seeing as memory safety of the module is tied to the context?

vosen:
Not sure if I follow. The `Module` type returned from `to_llvm_module` should contain both `Context` from `emit_llvm.rs` and `Module` from `emit_llvm.rs` that were created inside `emit_llvm::run`.
As for drop order, rust has a defined drop order: "The fields of a [struct](https://doc.rust-lang.org/reference/types/struct.html) are dropped in declaration order.". More here: https://doc.rust-lang.org/reference/destructors.html#destructors
Does it answer your question?

JoelleJS:
@vosen I wasn't passing the `Context` at first because it wasn't used at any point, but I started getting memory access errors in my tests. When I did include it in the output of `emit_llvm::run` and include it in `pass::Module` this issue was fixed. So it seems that the validity of the `LLVMModuleRef` depends on the `LLVMContextRef` not being dropped earlier.

I'm probably not telling you anything you don't know up to this point. What I am suggesting is to include the `emit_llvm::Context` _in_ `emit_llvm::Module`, as the second is dependent on the first or else memory issues start to occur. Leaving the responsibility of keeping the `Context` in scope to the caller of `emit_llvm::run`, a supposedly safe function, in order to keep the `Module` valid seems unsafe. By including the `emit_llvm::Context` in the `emit_llvm::Module` we would counter this unsafety as soon as possible.

JoelleJS:
For the sake of example I have committed the change that I'd like to make. It can of course be reverted if you don't agree with it.

PR #342, opened 2025-02-22
Implement mode tracking for AMD GPU

### LINKED ISSUE #289: Handle ftz flag correctly
Expected outcome:
All instructions with .ftz modifiers correctly flush subnormals (or not, if the flag is unset). Even in the modules with mixed ftz-on/ftz-off instructions.

Comments:
* Most of the time a module will either have all instructions ftz-on or all instructions ftz-off. Implementation should make use of that fact and visit all instructions if we are in such case. Consult with with LLVM, I think AMDGPU LLVM supports ftz attribute function-wide
* If it's mixed ftz mode then best course of action is to set ftz off and flush per-argument. `ptx_parser` and its visitors (`visit`/`visit_mut`/`visit_map`) should be extended to report ftz information.   Changes to `generate_instruction_type!` will also be necessary to expose per-argument ftz information.
* `llvm.canonicalize` can probably do the flushing, needs checking

vosen:
This is not done yet, but I believe the bulk of the work is done
Will close #289 once merged


vosen:
> I believe the bulk of the work is done

So that was so wrong. It's finished now, but I'll leave it up for few hours in case of any last second thoughts

Issue #284, opened 2024-10-13
Add offline compiler


vosen:
Expected outcome:
A new subproject in ZLUDA: a command line tool zoc.exe (ZLUDA offline compiler) that compiles PTX to the final binary and saves it to disk along with intermediate formats

Comments:
* It's a developer's tool, we don't expect it to be used by the end-users
* For consistency with related projects (https://github.com/vosen/ptx_tests) it should use bpaf for command line (https://crates.io/crates/bpaf)
* It should take PTX and emit 
  * Pre-linked LLVM bitcode texts (.ll)
  * Post-linking LLVM bitcode text (.ll) (optional if difficult)
  * Final ELF binary (.elf)
  * Disassembly of the final ELF binary (.asm) (optional if difficult)
* For implementation, look at how ZLUDA uses comgr and refactor it to save intermediates

JoelleJS:
I've started (slowly) working on this

vosen:
Closed as of #344

PR #344, opened 2025-03-03
zoc (ZLUDA offline compiler)

JoelleJS:
Implements #284

_It should take PTX and emit_
- [x] _Pre-linked LLVM bitcode texts (.ll)_ 
- [x] _Post-linking LLVM bitcode text (.ll) (optional if difficult)_
- [x] _Final ELF binary (.elf)_
- [x] _Disassembly of the final ELF binary (.asm) (optional if difficult)_

Todos:
- [x] Add option to specify output path
- [x] Test if error messages make sense


JoelleJS:
@vosen I have a working version that is ready for testing and feedback (will probably be making some small changes still but the functionality is there)

Example commands:
```
./target/debug/zoc --ll ptx/src/test/spirv_run/add.ptx
./target/debug/zoc --linked-ll ptx/src/test/spirv_run/add.ptx
./target/debug/zoc --elf ptx/src/test/spirv_run/add.ptx
./target/debug/zoc --asm ptx/src/test/spirv_run/add.ptx
```

compiler/src/main.rs
```diff
@@ -0,0 +1,208 @@
...
+    /// Output path
+    output_path: Option<PathBuf>,
+
+    #[bpaf(positional("filename"))]
+    /// PTX file
+    ptx_path: String,
+}
+
+fn main() -> ExitCode {
+    main_core().map_or_else(
```
vosen: `map_or` would be minimally simpler here

compiler/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+path = "src/main.rs"
+
+[dependencies]
+amd_comgr-sys = { path = "../ext/amd_comgr-sys" }
+bpaf = { version = "0.9.19", features = ["derive"] }
+comgr = { path = "../comgr" }
+hip_runtime-sys = { path = "../ext/hip_runtime-sys" }
+ptx = { path = "../ptx" }
+ptx_parser = { path = "../ptx_parser" }
+thiserror = "2.0.12"
```
vosen: Needs
```
[package.metadata.zluda]
debug_only = true
```
so it is part of xtask compilation

compiler/src/main.rs
```diff
@@ -0,0 +1,208 @@
...
+}
+
+#[derive(Debug)]
+struct LLVMArtifacts {
+    bitcode: Vec<u8>,
+    linked_bitcode: Vec<u8>,
+    llvm_ir: Vec<u8>,
+}
+
+fn get_arch() -> Result<CString, CompilerError> {
```
vosen: It's absolutely a valid flow to compile without AMD GPU or compile for a GPU you don't have installed.
This is a good default, but we need some more:
* An option to pass the architecture name to compile for
* If any of the hip runtime functions to detect gpu fails (or there is no gpu) then use a default: gfx1100

compiler/src/main.rs
```diff
@@ -0,0 +1,208 @@
...
+    use hip_runtime_sys::*;
+    unsafe { hipInit(0) }?;
+    let mut dev_props: MaybeUninit<hipDeviceProp_tR0600> = MaybeUninit::uninit();
```
vosen: I prefer this pattern:
```
let mut dev_props = unsafe { mem::zeroed() };
unsafe { hipGetDevicePropertiesR0600(&mut dev_props, 0) }?;
```
All HIP objects are plain data, zero-initializing them is fine. And one does not need to remember to call assume_init()

compiler/src/main.rs
```diff
@@ -0,0 +1,208 @@
...
+    unsafe { hipGetDevicePropertiesR0600(dev_props.as_mut_ptr(), 0) }?;
+    let dev_props = unsafe { dev_props.assume_init() };
+    let arch = dev_props.gcnArchName;
+    let arch: Vec<u8> = arch
```
vosen: You can just `CStr::from_ptr(dev_props.gcnArchName.as_ptr())` or `CStr::from_bytes_until_nul(&dev_props.gcnArchName)` if you watnt to check

compiler/src/main.rs
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    let ptx_path = Path::new(&opts.ptx_path).to_path_buf();
+    check_path(&ptx_path)?;
+
+    let output_path = match opts.output_path {
+        Some(value) => value,
+        None => get_output_path(&ptx_path, &output_type)?,
+    };
+    check_path(&output_path)?;
```
vosen: No reason to check output path. We should just create it (and if it fails during creation then expose that failure)

compiler/src/main.rs
```diff
@@ -0,0 +1,208 @@
...
+        let error = CompilerError::GenericError {
+            cause: None,
+            message,
+        };
+        return Err(error);
+    }
+    Ok(())
+}
+
+fn get_output_path(ptx_path: &PathBuf, output_type: &OutputType) -> Result<PathBuf, CompilerError> {
```
vosen: By default we should build in the folder where .ptx is located. That's because the default use is you are going to do multiple runs of `zluda_dump` which will write out all teh PTX to separate directories. So you will have multiple ptx from different runs or and different applications: `C:\temp\zluda\geekbench_x86_64.exe\module_0001_05.ptx`, `C:\temp\zluda\geekbench_x86_64.exe_1\module_0001_05.ptx`, `C:\temp\zluda\python.exe\module_0001_05.ptx`.
It makes it easier to then navigate if the zoc output is then right next to ptx

compiler/src/main.rs
```diff
@@ -0,0 +1,208 @@
...
+    let opts = options().run();
+
+    let output_type = opts.output_type.unwrap_or_default();
+
+    if opts.linked && output_type != OutputType::LlvmIr {
+        println!("Warning: option --linked only makes sense when combined with --ll. Ignoring.");
+    }
+
+    let ptx_path = Path::new(&opts.ptx_path).to_path_buf();
+    check_path(&ptx_path)?;
```
vosen: Don't check the path. Actually opening a file is enough of a check. And logically you can have a situation where your check passes, but file can't be opened anyway. Separate file open check is completely superfluous

vosen (changes requested):
Good changes. Other than the things I directly commented on there's one major thing I want: simplification of output types. Instead of having a flag for each separate output type just output all of them.
Usage of `zoc` is usually exploratory: something fails at runtime or during compilation or is slow and you might check suspecting a certain stage of the compilation process, but quite often you will want to check other stages too.
Just make sure to output file before moving to the next stage. All comgr calls are calls to llvm and llvm can kill the process in a case of failure. THe trickiest part is probably going between various stages of bitcode linking. This can be done with `AMD_COMGR_ACTION_COMPILE_SOURCE_WITH_DEVICE_LIBS_TO_BC`, `AMD_COMGR_ACTION_CODEGEN_BC_TO_RELOCATABLE`, `AMD_COMGR_ACTION_LINK_RELOCATABLE_TO_EXECUTABLE`. You can check history of comgr crate, that's how it used to work before I realized it can be simplified

JoelleJS:
@vosen Quickly checking in to say that I'm still alive and will pick this up hopefully soon, regular life has been busy the last month

vosen:
Great. In the meantime I've tried to restrain myself with messing too much with comgr code, but did not succeed :)

JoelleJS:
Got a bit discouraged by the merge conflicts back when I looked at this, so I'll have to look further into what has changed in master since I built this and see if a tricky merge or just salvaging code from this branch and putting it in a new branch is the quickest approach. Have two weeks of holiday as of next Wednesday and I really want to finish this because it's a handy tool that's been delayed for months.

compiler/src/main.rs
JoelleJS: `map_or` executes a function on success or returns a value on failure. We need the other way around here. But I have chosen another structure that doesn't use a map function.

compiler/src/main.rs
JoelleJS: There's now a default arch `gfx1000` and an `--arch` option :)

JoelleJS:
> Good changes. Other than the things I directly commented on there's one major thing I want: simplification of output types. Instead of having a flag for each separate output type just output all of them. Usage of `zoc` is usually exploratory: something fails at runtime or during compilation or is slow and you might check suspecting a certain stage of the compilation process, but quite often you will want to check other stages too. Just make sure to output file before moving to the next stage. All comgr calls are calls to llvm and llvm can kill the process in a case of failure. THe trickiest part is probably going between various stages of bitcode linking. This can be done with `AMD_COMGR_ACTION_COMPILE_SOURCE_WITH_DEVICE_LIBS_TO_BC`, `AMD_COMGR_ACTION_CODEGEN_BC_TO_RELOCATABLE`, `AMD_COMGR_ACTION_LINK_RELOCATABLE_TO_EXECUTABLE`. You can check history of comgr crate, that's how it used to work before I realized it can be simplified

Okay, this definitely requires some changes but let me have a look

JoelleJS:
Okay, I believe I have done everything except the following,
> THe trickiest part is probably going between various stages of bitcode linking. This can be done with AMD_COMGR_ACTION_COMPILE_SOURCE_WITH_DEVICE_LIBS_TO_BC, AMD_COMGR_ACTION_CODEGEN_BC_TO_RELOCATABLE, AMD_COMGR_ACTION_LINK_RELOCATABLE_TO_EXECUTABLE. You can check history of comgr crate, that's how it used to work before I realized it can be simplified

I would like to propose that we keep this split of `AMD_COMGR_ACTION_COMPILE_SOURCE_TO_EXECUTABLE` into three parts out of the scope of this MR, because I would really like to get this merged while it's up to date with the master branch 😅

Changes:

- The compiler now outputs all stages by means of a hook into `comgr::compile_bitcode`
- The output directory is the same as the PTX file by default, but can be changed using `--output-dir`
- The architecture is chosen from the highest to the lowest priority: the one specified using `--arch`; the one detected from your GPU if possible; the default value `gfx1100`

JoelleJS:
@vosen @zluda-violet Ready for review

vosen (approved):

vosen:
Thank you, I'm very happy with the changes

PR #351, opened 2025-03-30
Implement mul24

JoelleJS:
See title.

Also I took the liberty to sort the instructions in `ast.rs` alphabetically. Can revert if you don't like it. I'd like to implement more instructions after this but had difficulty seeing which instructions are and aren't implemented.


ptx/src/pass/emit_llvm.rs
```diff
@@ -2281,15 +2282,51 @@ impl<'a> MethodEmitContext<'a> {
...
-        self.emit_intrinsic(
-            c"llvm.amdgcn.mul.u24",
+        let name_lo = match data.type_ {
+            ast::ScalarType::U32 => c"llvm.amdgcn.mul.u24",
+            ast::ScalarType::S32 => c"llvm.amdgcn.mul.i24",
+            _ => return Err(error_unreachable()),
+        };
+        let res_lo = self.emit_intrinsic(
+            name_lo,
             Some(arguments.dst),
```
vosen: This should be `None` if it is the `.hi` variant. `Some(x)` means "emit intrinsic and register returned `LLVMValueRef` to `x`", `None` means "emit instrinsic and return `LLVMValueRef`, the returned value does not correspond to any ZLUDA identifier - it's some intermediate LLVM value".
It all works right now somewhat by accident because in `.hi` case we override the value down below, but strictly speaking goes against invariants. I should check if we are not overriding things in `ResolveIdent::register` (at least in debug)

ptx/src/pass/emit_llvm.rs
```diff
@@ -2281,15 +2282,51 @@ impl<'a> MethodEmitContext<'a> {
...
+                name_hi,
+                None,
+                Some(&ast::Type::Scalar(data.type_)),
+                vec![
+                    (src1, get_scalar_type(self.context, data.type_)),
+                    (src2, get_scalar_type(self.context, data.type_)),
+                ],
+            )?;
+            let shift_number = unsafe { LLVMConstInt(LLVMInt32TypeInContext(self.context), 16, 0) };
+            let res_lo_shr = unsafe {
```
vosen, line 2319: Use `LLVM_UNNAMED.as_ptr()` instead of proper names. While LLVM is nice enough to rename multiple declarations, I think it can still accidentally rename kernels if they are named "res_lo_shr" and come after this variable

ptx/src/pass/emit_llvm.rs
```diff
@@ -2281,15 +2282,51 @@ impl<'a> MethodEmitContext<'a> {
...
+                LLVMBuildLShr(self.builder, res_lo, shift_number, c"res_lo_shr".as_ptr())
+            };
+            let res_hi_shl =
+                unsafe { LLVMBuildShl(self.builder, res_hi, shift_number, c"res_hi_shl".as_ptr()) };
+
+            self.resolver
+                .with_result(arguments.dst, |dst: *const i8| unsafe {
+                    LLVMBuildAdd(self.builder, res_lo_shr, res_hi_shl, dst)
```
vosen: Please use `LLVMBuildOr` so it's a bit more obvious that we are doing bit operations here

vosen (changes requested):
Mostly good, but I want minor improvements for robustness

JoelleJS:
@vosen I believe the last commit addresses all your comments

vosen:
👍

PR #356, opened 2025-04-03
Update README.md

adalava:
List AMD HIP64 library as dependency

On clean Ubuntu 24.04 system I had to install libamdhip64-dev to fix  `/usr/bin/ld: cannot find -lamdhip64: No such file or directory` during build.


vosen:
Thanks, added in #474 as part of a bigger docs rewrite

PR #374, opened 2025-05-26
[WIP] Start working on PhysX 32bit

vosen:


vosen:
I've tried a bunch of things with zluda_dump. There's some brokenness with how we handle dark API. The best log I got was after changing dark apis from "system" to "C" ABI: [log.txt](https://github.com/user-attachments/files/20446585/log.txt). With this the physx auto-downgraded into cpu mode.

It's a combination of:
* Some dark APIs have incorrect signatures
* They all use "C" ABI (our thunk messes up the stack)
* Some other factor, because changing `get_export_override_fn` to just return `original_fn` did not solve the issue

The path forward is to manually (without `zluda_dump`) check every dark API used by fluidmark and decompile again and correct the signatures.

vosen:
Rustup command: `rustup install stable-i686-pc-windows-msvc`
Build command: `cargo +stable-i686-pc-windows-msvc build -p zluda_inject -p zluda_dump -p zluda_redirect`

vosen:
This mostly allows one to to produce a log, but it requires disabling nvml injection. Log attached.
Overall `zluda_injector` is not in a great shape and requires a bunch of improvements:
* There should be a common trait/macro to handle each of nvml/nvcuda/perf libs. They behave pretty much the same, just have different names. Both on the injector side and redirector side
* I am fairly sure that redirector injects incorrectly into child processes (utf16 instead of utf8)
* Another injection method (launch remote thread) for cases where current injection won't work (non-ascii paths)
* Pretty sure we leak temporary nvcuda.dll
* Would be nice to support mixed 32/64 bit injections

[FluidMark.exe.zip](https://github.com/user-attachments/files/21113722/FluidMark.exe.zip)

PR #377, opened 2025-06-10
Implement cuGetProcAddress and cuGetProcAddress_v2

zluda-violet:
This is mostly the same implementation used by zluda_dump.

### DIFF
--- a/zluda/src/impl/driver.rs
+++ b/zluda/src/impl/driver.rs
@@ -5,8 +5,9 @@ use cuda_types::cuda::*;
 use hip_runtime_sys::*;
 use std::{
     ffi::{CStr, CString},
-    mem, slice,
+    mem, ptr, slice,
     sync::OnceLock,
+    usize,
 };
 
 pub(crate) struct GlobalState {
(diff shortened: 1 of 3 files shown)


zluda/src/impl/driver.rs
```diff
@@ -79,3 +80,58 @@ pub(crate) fn get_version(version: &mut ::core::ffi::c_int) -> CUresult {
...
+    symbol: *const ::core::ffi::c_char,
+    pfn: *mut *mut ::core::ffi::c_void,
+    cuda_version: ::core::ffi::c_int,
+    flags: cuda_types::cuda::cuuint64_t,
+) -> CUresult {
+    get_proc_address_v2(symbol, pfn, cuda_version, flags, None)
+}
+
+pub(crate) unsafe fn get_proc_address_v2(
+    symbol: *const ::core::ffi::c_char,
```
vosen: It's not safe enough, add a `FromCuda` impl that converts from `*const c_char` to `CStr` (and returns INVALID_VALUE if it's a null ptr)

zluda/src/impl/driver.rs
```diff
@@ -79,3 +80,58 @@ pub(crate) fn get_version(version: &mut ::core::ffi::c_int) -> CUresult {
...
+    pfn: *mut *mut ::core::ffi::c_void,
```
vosen: This should be `&mut *mut ::core::ffi::c_void`, if necessary, extend `FromCuda`

PR #380, opened 2025-06-11
Implement runtime_callback_hooks_fn2

zluda-violet:
This starts the implementation of cuGetExportTable and the dark API. The only function implemented at this point is runtime_callback_hooks_fn2. I'm not sure about the correctness of returning a mutable reference to a non-mutable static variable, but the compiler allows it.

### DIFF
--- a/zluda/src/impl/driver.rs
+++ b/zluda/src/impl/driver.rs
@@ -77,7 +77,20 @@ pub(crate) fn init(flags: ::core::ffi::c_uint) -> CUresult {
 }
 
 const FN2_BUFFER_SIZE: usize = 400;
-static FN2_BUFFER: [u8;FN2_BUFFER_SIZE] = [0;FN2_BUFFER_SIZE];
+
+struct Fn2Buffer {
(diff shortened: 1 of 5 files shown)


zluda/src/impl/driver.rs
```diff
@@ -76,6 +76,183 @@ pub(crate) fn init(flags: ::core::ffi::c_uint) -> CUresult {
     Ok(())
 }
 
+const FN2_BUFFER_SIZE: usize = 400;
```
jiridanek, line 79: `400` did not work for me; after I changed this to 1024, same as

https://github.com/pole5tar/ZLUDA-fork/blob/4a81dbffb5ce37be0f999476ce69c444d9ec77c8/zluda/src/impl/dark_api.rs#L38-L39

has that, program (just a basic `cudaGetDeviceCount`) ~stopped segfaulting for me~ still segfaults, but later and hopefully for different reasons;

I have heavily modified ZLUDA, so the observation may not be applicable directly; just wanted to give hint to question the size of the buffer.

zluda-violet: You're correct, thanks for catching this.

PR #387, opened 2025-06-17
Use `integrity_check` implementation

zluda-violet:
This handles the dark API integrity check using the existing implementation. Also implements `check_fn3` by simply returning 0 – this seems to work for now.


vosen:
looks good, I'll merge once builds pass

PR #388, opened 2025-06-18
Implement `cuLibraryLoadData`

zluda-violet:
Adds an implementation for `cuLibraryLoadData`. For now, libraries are loaded the same as modules, and the jit and library options are ignored. This change also adds support for modules that are provided as a fatbin rather than as a null-terminated PTX string.


cuda_types/src/dark_api.rs
```diff
@@ -197,3 +211,43 @@ impl ParseError {
...
+                return Ok(decompressed_vec);
+            }
+        }
+    }
+}
+
+pub unsafe fn decompress_zstd(file: &FatbinFileHeader) -> Result<Vec<u8>, ParseError> {
+    let mut result = Vec::with_capacity(file.uncompressed_payload as usize);
+    let payload = file.get_payload();
+    dbg!((payload.len(), file.uncompressed_payload, file.payload_size));
```
vosen: Oops, I did not realize I left that in, remove it

cuda_types/src/dark_api.rs
```diff
@@ -197,3 +211,43 @@ impl ParseError {
         }
     }
 }
+
+const MAX_MODULE_DECOMPRESSION_BOUND: usize = 64 * 1024 * 1024;
+
+pub unsafe fn decompress_lz4(file: &FatbinFileHeader) -> Result<Vec<u8>, ParseError> {
```
vosen: Move decompression code to `dark_api` crate. I want cuda_types to contain just type declarations. the code in this file is an exception for convenience, but maybe it too should move to dark api crate?

zluda/src/impl/module.rs
```diff
@@ -18,12 +21,47 @@ impl ZludaObject for Module {
     }
 }
 
-pub(crate) fn load_data(module: &mut CUmodule, image: *const std::ffi::c_void) -> CUresult {
+fn get_ptx_from_wrapped_fatbin(image: *const ::core::ffi::c_void) -> Result<Vec<u8>, CUerror> {
```
vosen, line 25: This is ok if it works with `llm.c`, but eventually we will want a more complex solution. Principally, a fatbin can contain multiple `.ptx` modules for different hardware architectures. ZLUDA itself will want to pick just one (I think a good approach is to pick the highest hardware version module that we manage to compile). zluda_dump will want to go through all ptx (and elf) in fatbin to save them to disk.
So in the future we will have common code that for a `FatbincWrapper` or `FatbinHeader` will give you iterator that goes through all elf/ptx modules.
Current approach is ok, but consider if you want to start adjusting code for that

zluda/src/impl/library.rs
```diff
@@ -0,0 +1,37 @@
...
+    base: hipModule_t,
+}
+
+impl ZludaObject for Library {
+    const COOKIE: usize = 0xb328a916cc234d7c;
+
+    type CudaHandle = CUlibrary;
+
+    fn drop_checked(&mut self) -> CUresult {
+        unsafe { hipModuleUnload(self.base) }?;
```
vosen, line 19: It will explode when I do `cuModuleUnload` and then `cuLibraryUnload`? I think doing `cuModuleUnload` on module that comes form a library is not allowed.
It's not important right now, but we will want to go back to this. Mark it with TODO for now. We will write tests that check for this later

cuda_types/src/dark_api.rs
zluda-violet: Done.

cuda_types/src/dark_api.rs
zluda-violet: I've refactored so that this file only has type declarations.

zluda/src/impl/module.rs
zluda-violet: I've added a comment noting this. I think the current structure shouldn't be difficult to change to support this when we need it.

zluda/src/impl/library.rs
zluda-violet: I've added the TODO.

PR #394, opened 2025-06-24
Check LLVM IR for `test_ptx!` with no input/output

zluda-violet:


ptx/src/test/spirv_run/mod.rs
```diff
@@ -243,7 +252,8 @@ test_ptx!(
 test_ptx!(assertfail);
 test_ptx!(func_ptr);
 test_ptx!(lanemask_lt);
-test_ptx!(extern_func);
```
vosen, line 246: I did not run it, but isn't there a typo somewhere? `extern_func` is commented out and has LLVM bitcode, while `func_ptr` is included but has no LLVM bitcode

vosen (changes requested):

ptx/src/test/spirv_run/mod.rs
zluda-violet: Yes, you're correct, thanks for catching that!

PR #395, opened 2025-06-25
Read test files at runtime for development ergonomics

zluda-violet:
Adds the `ptx/ci_build` feature to enable test files to be bundled directly in the executable.

### DIFF
--- a/ptx/src/test/spirv_run/mod.rs
+++ b/ptx/src/test/spirv_run/mod.rs
@@ -20,13 +20,11 @@ macro_rules! read_test_file {
         {
             // CARGO_MANIFEST_DIR is the crate directory (ptx), but file! is relative to the workspace root (and therefore also includes ptx). 
             let mut path = PathBuf::new();
-            println!("{}", file!());
             path.push(env!("CARGO_MANIFEST_DIR"));
             path.pop();
             path.push(file!());
             path.pop();
             path.push($file);
-            println!("{:?}", path);
             std::fs::read_to_string(path).unwrap()
         }
     };
--- a/ptx/src/test/spirv_run/mod.rs
+++ b/ptx/src/test/spirv_run/mod.rs
@@ -20,11 +20,10 @@ macro_rules! read_test_file {
         {
             // CARGO_MANIFEST_DIR is the crate directory (ptx), but file! is relative to the workspace root (and therefore also includes ptx). 
             let mut path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
-
-            let mut file_dir = PathBuf::from(file!());
-            file_dir.pop();
-
-            path.extend(["..", file_dir.to_str().unwrap(), $file]);
+            path.pop();
+            path.push(file!());
+            path.pop();
+            path.push($file);
             std::fs::read_to_string(path).unwrap()
         }
     };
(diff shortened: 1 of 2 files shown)


ptx/src/test/spirv_run/mod.rs
```diff
@@ -10,38 +10,63 @@ use std::fmt::{self, Debug, Display, Formatter};
...
 use std::ptr;
 use std::str;
 
+#[cfg(not(feature = "ci_build"))]
+macro_rules! read_test_file {
+    ($file:expr) => {
+        {
+            // CARGO_MANIFEST_DIR is the crate directory (ptx), but file! is relative to the workspace root (and therefore also includes ptx). 
+            let mut path = PathBuf::new();
+            println!("{}", file!());
```
vosen: This `println!` and the next one are debugging leftovers?

ptx/src/test/spirv_run/mod.rs
```diff
@@ -10,38 +10,63 @@ use std::fmt::{self, Debug, Display, Formatter};
...
+            path.push(env!("CARGO_MANIFEST_DIR"));
+            path.pop();
```
vosen, line 23: The pop and push sequence seems to be superfluous (push an then pop) in some ways? And can be replaced with `.extend([ ... ])`

vosen (changes requested):

ptx/src/test/spirv_run/mod.rs
zluda-violet: Yes, thanks for catching that. I've removed them.

ptx/src/test/spirv_run/mod.rs
zluda-violet: `PathBuf` has a different interface and semantics than `Vec`. [`push`](https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.push) behaves similarly to `extend`, and that's how it's being used here. It's extending the path by another path. `pop`, on the other hand, is only removing the final component of the `PathBuf`.

To show what's happening for `ld_st_implicit.ptx`:

```
path.push(env!("CARGO_MANIFEST_DIR"));
    => /home/violet/ZLUDA/ptx
path.pop();
    => /home/violet/ZLUDA
path.push(file!());
    => /home/violet/ZLUDA/ptx/src/test/spirv_run/mod.rs
path.pop();
    => /home/violet/ZLUDA/ptx/src/test/spirv_run
path.push($file);
    => /home/violet/ZLUDA/ptx/src/test/spirv_run/ld_st_implicit.ptx
```

vosen: Ah, I did not realize that. Still, how about making this more compact like this:
```
    let mut path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
    path.extend(["..", file!(), "..", $file]);
```

zluda-violet: `file!()` ends with the filename, and we get an OS error about it not being a directory if we use `..` after it, so we have to use `pop()`. I've changed it so that it uses `pop` just for `file!()`, and `extend` otherwise, but I'm not sure it's much more readable.

vosen: Hmm, it could be `.extend(["..", PathBuf::from(file!()).parent().unwrap().to_str().unwrap(), $file])`, but that's just as bad.
Please revert to the original push and pop sequence, that was much clearer

zluda-violet: Done.

PR #397, opened 2025-06-27
Make `derive_parser` work with all optional arguments

zluda-violet:
The current implementation using `winnow`'s `opt` does not work for optional arguments that are in the middle of the command. For example,

```
bar{.cta}.red.op.pred   p, a{, b}, {!}c;
```

This is because `opt` is greedy, and will always match `{, b}` instead of `,{!} c`. This change switches to using `alt` in this situation.

The `alt` implementation will also be used for `[|q]` patterns, since those are marked as optional. I believe this will still work, even though not strictly necessary.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -605,6 +605,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "js-sys"
+version = "0.3.77"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1cfaf33c695fc6e08064efbc1f72ec937429614f25eef83af942d0e227c3a28f"
+dependencies = [
+ "once_cell",
+ "wasm-bindgen",
+]
+
+[[package]]
 name = "lazy_static"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


vosen:
This is good first attempt, but I think this can be made cleaner if we take a step back and frame the problem differently:
There's two separate concerns:
* How to parse a single argument with all its surrounding sigils: `[`, `|`
* How to _combine_ the parser for an argument with the remaining parsers, so in the `bar` example how to combine `a` with `b` and `c`

With this formulation there are two possible combinators: sequence (winnow tuples) or optional (all that complicated stuff around `alt` in the current code). Main problem with optionals that leads to all that extra gymnastics is the fact that we don't have access to right side parsers, they remain to be parsed, hence the recursive `emit_argument_parser`.

I think this can be avoided by generating the argument parser right-to-left: do an `rfold` on `definition.arguments.0`, your accumulator is going to be the TokenStream for right side arguments, you compute TokenStream for parser of the current argument and then join them. If it's a non-optional then do a normal tuple, if it's an optional then I think it's best to write (and use) our own combinator that takes two parsers `p1`, `p2` and whose logic is basically `alt(((empty, p2), (p1, p2)))`, this way you should be able to avoid generating `parser_...` fns and sidestep type checking issues.

At the end you'd unpack values to pass to the function `bar_0` from a parser in our `bar` example like this: `let (p, (a, (b, c))) = ...`

zluda-violet:
@vosen updated according to your suggestions

PR #398, opened 2025-06-27
Unified fatbin versions behind a single iterator.

aiwhskruht:
This change fixes a bug in fatbin loading that didn't respect mutual exclusion for versions. 

Tested with a bunch of cuda-samples through zluda_dump with and without zluda/HIP.


dark_api/src/fatbin.rs
```diff
@@ -107,12 +103,33 @@ impl<'a> FatbinSubmodule<'a> {
...
+                    Err(_) => Ok(None),
+                }
+            },
+        }
+    }
+}
+
+pub struct FatbinSubmoduleIterator<'a> {
     fatbins: *const *const std::ffi::c_void,
+    _phantom: std::marker::PhantomData<&'a FatbinHeader>,
```
aiwhskruht, line 129: We could try to avoid lifetimes here entirely, but since Fatbin is already tagged this seems better.

vosen: It's a good change, previous code was wrong. In real life all those objects have 'static lifetimes, but we should only rely on thos eobjects being alive for the scope of the function and those lifetime enforce it

dark_api/src/fatbin.rs
```diff
@@ -75,21 +75,17 @@ impl<'a> Fatbin<'a> {
...
-            parse_fatbin_header(&self.wrapper.data).map_err(|e| FatbinError::ParseFailure(e))?;
-        Ok(FatbinSubmodule::new(header))
-    }
-
-    pub fn get_submodules(&self) -> Option<FatbinSubmoduleIterator> {
-        let is_version_2 = self.wrapper.version == FatbincWrapper::VERSION_V2;
-        if !is_version_2 {
-            return None;
+    pub fn get_submodules(&self) -> Result<FatbinIter<'a>, FatbinError> {
+        if self.wrapper.version == FatbincWrapper::VERSION_V2 {
```
vosen: Please explictly check if it's v1 (and do FatbinIter::V1) or v2 (and do FatbinIter::V2) and if it's not then fail

dark_api/src/fatbin.rs
```diff
@@ -107,12 +103,33 @@ impl<'a> FatbinSubmodule<'a> {
...
+
+impl<'a> FatbinIter<'a> {
+    pub fn next(&mut self) -> Result<Option<FatbinSubmodule<'a>>, ParseError> {
+        match self {
+            FatbinIter::V1(opt) => Ok(opt.take()),
+            FatbinIter::V2(iter) => unsafe {
+                match iter.next() {
+                    Ok(Some(submodule)) => Ok(Some(submodule)),
+                    Ok(None) => Ok(None),
+                    Err(_) => Ok(None),
```
vosen: Why swallow the error here?

zluda/src/impl/module.rs
```diff
@@ -24,14 +24,15 @@ impl ZludaObject for Module {
...
 
-    while let Some(file) = unsafe { files.next().map_err(|_| CUerror::UNKNOWN)? } {
-        // Eventually we will want to get the PTX for the highest hardware version that we can get to compile. But for now we just get the first PTX we can find.
-        if file.header.kind == FatbinFileHeader::HEADER_KIND_PTX {
-            let decompressed = unsafe { file.decompress() }.map_err(|_| CUerror::UNKNOWN)?;
-            return Ok(decompressed);
+    while let Some(current) = unsafe { submodules.next().map_err(|_| CUerror::UNKNOWN)? } { 
+        let mut files = current.get_files();
+        while let Some(file) = unsafe { files.next().map_err(|_| CUerror::UNKNOWN)? } {
+            if file.header.kind == FatbinFileHeader::HEADER_KIND_PTX && file.header.payload_size > 0 {
```
vosen: This zero size check should not be necessary, are there cases where it helps?

aiwhskruht: Good point, this was left over from some checks I did before making V1 and V2 paths separate. This should never happen now since it was only an issue when we were dumping the empty V1 headers even for V2. I'll remove it.

dark_api/src/fatbin.rs
aiwhskruht: Sure, I actually had that but removed it because of the coverage in parse_fatbinc_wrapper(). I'll use a match here with an default error.

dark_api/src/fatbin.rs
```diff
@@ -107,12 +107,33 @@ impl<'a> FatbinSubmodule<'a> {
...
+    V1(Option<FatbinSubmodule<'a>>),
+    V2(FatbinSubmoduleIterator<'a>),
+}
+
+impl<'a> FatbinIter<'a> {
+    pub fn next(&mut self) -> Result<Option<FatbinSubmodule<'a>>, ParseError> {
+        match self {
+            FatbinIter::V1(opt) => Ok(opt.take()),
+            FatbinIter::V2(iter) => unsafe {
+                match iter.next() {
```
vosen: One last question: can't this whole block can be just `iter.next()`?

aiwhskruht: Indeed. I'll update.

aiwhskruht: Done.

PR #400, opened 2025-07-02
Add warp-wide tests

zluda-violet:
Adds `test_ptx_warp!`, along with a simple test to check that it works correctly.

### DIFF
--- a/ptx/src/test/spirv_run/mod.rs
+++ b/ptx/src/test/spirv_run/mod.rs
@@ -36,28 +36,8 @@ macro_rules! read_test_file {
     };
 }
 
-macro_rules! test_ptx {
-    ($fn_name:ident, $input:expr, $output:expr) => {
-        paste::item! {
-            #[test]
-            fn [<$fn_name _hip>]() -> Result<(), Box<dyn std::error::Error>> {
-                let ptx = read_test_file!(concat!(stringify!($fn_name), ".ptx"));
-                let input = $input;
-                let mut output = $output;
-                test_hip_assert(stringify!($fn_name), &ptx, &input, &mut output)
-            }
-        }
-
-        paste::item! {
-            #[test]
-            fn [<$fn_name _cuda>]() -> Result<(), Box<dyn std::error::Error>> {
-                let ptx = read_test_file!(concat!(stringify!($fn_name), ".ptx"));
-                let input = $input;
-                let mut output = $output;
-                test_cuda_assert(stringify!($fn_name), &ptx, &input, &mut output)
-            }
-        }
-
+macro_rules! test_ptx_llvm {
+    ($fn_name:ident) => {
         paste::item! {
             #[test]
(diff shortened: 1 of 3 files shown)


ptx/src/test/spirv_run/mod.rs
```diff
@@ -53,30 +66,40 @@ macro_rules! test_ptx {
...
+    ($fn_name:ident, $output:expr) => {
         paste::item! {
             #[test]
-            fn [<$fn_name _llvm>]() -> Result<(), Box<dyn std::error::Error>> {
+            fn [<$fn_name _hip>]() -> Result<(), Box<dyn std::error::Error>> {
                 let ptx = read_test_file!(concat!(stringify!($fn_name), ".ptx"));
-                let ll = read_test_file!(concat!("../ll/", stringify!($fn_name), ".ll"));
-                test_llvm_assert(stringify!($fn_name), &ptx, ll.trim())
+                let mut output = $output;
+                test_hip_assert(stringify!($fn_name), &ptx, None::<&[u8]>, &mut output, 64)
```
vosen, line 89: CUDA warp is 32 elements wide, not 64

ptx/src/test/spirv_run/mod.rs
```diff
@@ -53,30 +66,40 @@ macro_rules! test_ptx {
...
         paste::item! {
             #[test]
-            fn [<$fn_name _llvm>]() -> Result<(), Box<dyn std::error::Error>> {
-                let ptx = include_str!(concat!(stringify!($fn_name), ".ptx"));
-                let ll = include_str!(concat!("../ll/", stringify!($fn_name), ".ll")).trim();
-                test_llvm_assert(stringify!($fn_name), ptx, &ll)
+            fn [<$fn_name _cuda>]() -> Result<(), Box<dyn std::error::Error>> {
+                let ptx = read_test_file!(concat!(stringify!($fn_name), ".ptx"));
+                let mut output = $output;
+                test_cuda_assert(stringify!($fn_name), &ptx, None::<&[u8]>, &mut output, 64)
```
vosen, line 98: CUDA warp is 32 elements wide, not 64

ptx/src/test/spirv_run/mod.rs
zluda-violet: Could we still use a 64-element width, so that the output for both AMD and CUDA is the same, and just change the name to be more accurate?

vosen: Maybe I got the intention wrong: do you want to test output from two warps (in which case it's 64) or a single warp (in which case it should be 32)?
Both AMD RDNA gpus and NV gpus are warp 32 (AMD is configurable 32/64, but we always use 32)

zluda-violet: To ensure that some instructions behave correctly, we probably should test output from multiple warps. For example, the `barrier` instructions can operate across multiple warps.

vosen: Fair point

PR #405, opened 2025-07-07
Implement more CUDA driver API to enable simple cuda-samples

aiwhskruht:
This enables some simple samples to run (e.g. vectorAdd).


zluda/src/impl/context.rs
```diff
@@ -1,23 +1,83 @@
...
+
+    pub(crate) fn reset(&mut self) -> CUresult {
+        for (key, data) in self.storage.iter_mut() {
+            if let Some(_cb) = data.reset_cb {
+                _cb(data.handle, *key as *mut c_void, data.value as *mut c_void);
+            }
+        }
+        self.ref_count = 0;
+        self.flags = 0;
+        self.modules.clear();
```
vosen: This should explicitly destroy modules and surface error code (if any)

zluda/src/impl/context.rs
```diff
@@ -1,23 +1,83 @@
...
+            flags: 0,
+            modules: FxHashSet::default(),
+            storage: FxHashMap::default(),
+        }
+    }
+
+    pub(crate) fn reset(&mut self) -> CUresult {
+        for (key, data) in self.storage.iter_mut() {
+            if let Some(_cb) = data.reset_cb {
+                _cb(data.handle, *key as *mut c_void, data.value as *mut c_void);
```
vosen, line 45: Hmmm, I wonder how original CUDA uses CUcontext. Mutex is not recursive and it will explode on some calls. Can you leave a comment about this? something like todo: check what happens when recursively calling CUDA driver API

zluda/src/impl/module.rs
```diff
@@ -88,7 +88,9 @@ pub(crate) fn load_data(module: &mut CUmodule, image: &std::ffi::c_void) -> CUre
 }
 
 pub(crate) fn unload(hmod: CUmodule) -> CUresult {
+    println!("UNLOAD");
```
vosen: debugging leftover?

vosen (changes requested):

zluda/src/impl/context.rs
aiwhskruht: Sure, I was wondering the same. I can also run a test with tracing to see if there are API calls resulting from cuCtxReset.

zluda/src/impl/context.rs
```diff
@@ -1,23 +1,86 @@
...
+        for (key, data) in self.storage.iter_mut() {
+            if let Some(_cb) = data.reset_cb {
+                // TODO: check that these callbacks do not call into the CUDA driver
+                _cb(data.handle, *key as *mut c_void, data.value as *mut c_void);
+            }
+        }
+        self.ref_count = 0;
+        self.flags = 0;
+        for hmod in self.modules.drain() {
+            super::drop_checked::<module::Module>(hmod)?;
```
vosen: This is still a problem: first failing module will prevent all other drop_checked to run. It should either be a `.drain().fold()` that folds error codes or something along the lines:
```
let mut result = Ok(());
for hmod in self.modules.drain() {
    result = combine_errors(result, super::drop_checked::<module::Module>(hmod));
}
return result;
```

vosen (changes requested):

zluda/src/impl/context.rs
aiwhskruht: Ah yes, I have the fold version stashed. Will update.

aiwhskruht: The fold just returns the first error if one exists, while still dropping all modules.

PR #409, opened 2025-07-10
Add support for `shfl.sync.MODE.b32`

zluda-violet:
Adds support for

* `shfl.sync.up.b32`
* `shfl.sync.down.b32`
* `shfl.sync.bfly.b32`
* `shfl.sync.idx.b32`

This uses the AMD HIP `__shfl` functions to implement them. Note that `__shfl` is slightly higher level than these instructions – it takes the width instead of an options bitfield, and does not return whether the computed index is out of range. This PR decodes the options bitfield, and duplicates the index calculation logic in order to do a bounds check.

### DIFF
--- a/ptx/src/test/ll/shfl_sync_bfly_b32_pred.ll
+++ b/ptx/src/test/ll/shfl_sync_bfly_b32_pred.ll
@@ -1,4 +1,4 @@
-declare { i32, i1 } @__zluda_ptx_impl_shfl_sync_bfly_b32_pred(i32, i32, i32, i32) #0
+declare [2 x i32] @__zluda_ptx_impl_shfl_sync_bfly_b32_pred(i32, i32, i32, i32) #0
 
 declare i32 @__zluda_ptx_impl_sreg_tid(i8) #0
 
@@ -22,9 +22,10 @@ define amdgpu_kernel void @shfl_sync_bfly_b32_pred(ptr addrspace(4) byref(i64) %
 "40":                                             ; preds = %"39"
   store i32 %"33", ptr addrspace(5) %"45", align 4
   %"52" = load i32, ptr addrspace(5) %"45", align 4
-  %2 = call { i32, i1 } @__zluda_ptx_impl_shfl_sync_bfly_b32_pred(i32 %"52", i32 3, i32 31, i32 -1)
-  %"65" = extractvalue { i32, i1 } %2, 0
-  %"51" = extractvalue { i32, i1 } %2, 1
+  %2 = call [2 x i32] @__zluda_ptx_impl_shfl_sync_bfly_b32_pred(i32 %"52", i32 3, i32 31, i32 -1)
+  %"65" = extractvalue [2 x i32] %2, 0
+  %3 = extractvalue [2 x i32] %2, 1
+  %"51" = trunc i32 %3 to i1
   store i32 %"65", ptr addrspace(5) %"46", align 4
   store i1 %"51", ptr addrspace(5) %"47", align 1
   %"53" = load i1, ptr addrspace(5) %"47", align 1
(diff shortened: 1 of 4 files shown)


ptx/lib/zluda_ptx_impl.cpp
```diff
@@ -4,6 +4,8 @@
 
 #include <cstddef>
 #include <cstdint>
+
+#define HIP_ENABLE_WARP_SYNC_BUILTINS 1
```
vosen: I think that's not necessary anymore?

ptx/lib/zluda_ptx_impl.cpp
```diff
@@ -170,6 +172,54 @@ extern "C"
     BAR_RED_IMPL(and);
     BAR_RED_IMPL(or);
 
+    struct ShflSyncResult {
+        uint32_t output;
+        bool in_bounds;
+    };
+
+    // shfl.sync opts consists of two values, the endpoint ID and the subsection mask.
```
vosen: This comment is outdated

ptx/lib/zluda_ptx_impl.cpp
```diff
@@ -170,6 +172,54 @@ extern "C"
...
+    // 11000 in binary. This indicates that the two most significant bits in the 5-bit lane ID are
+    // the subsection address. For example, for a lane ID 13 (0b01101) the address of the beginning
+    // of the subsection is 0b01000 (8).
+    //
+    // The endpoint ID is the lane ID of the max lane ID for a specific mode. For the CUDA
+    // __shfl_sync intrinsics, it is always 31 for idx, bfly, and down, and 0 for up. For now, it
+    // seems like we don't need to use this.
+
+    #define SHFL_SYNC_IMPL(mode, fn, out_of_bounds_check)                                                                                     \
+    uint32_t FUNC(shfl_sync_##mode##_b32)(uint32_t input, int32_t idx, uint32_t opts, uint32_t membermask __attribute__((unused)))            \
```
vosen: Wouldn't it be simpler to write this function in terms of the `_pred` variant?

ptx/lib/zluda_ptx_impl.cpp
```diff
@@ -170,6 +172,54 @@ extern "C"
...
+    {                                                                                                                                         \
+        uint32_t section_mask = (opts >> 8) & 0b11111;                                                                                        \
+        uint32_t __attribute__((unused)) endpoint = opts & 0b11111;                                                                           \
```
vosen: Why ignore this value? It seems to me that it takes part in index calculations. I've tried one of the unit tests with opt changed from 4127 to 4096 and original CUDA produced a different result

vosen (changes requested):
BTW, but did you check if anonymous structs are linked by LLVM properly?

ptx/lib/zluda_ptx_impl.cpp
zluda-violet: Correct, removed.

ptx/lib/zluda_ptx_impl.cpp
zluda-violet: It is not; it is about the `shfl.sync` instruction that we are implementing, not the `__shfl` function we are calling.

ptx/lib/zluda_ptx_impl.cpp
zluda-violet: Yes, done.

ptx/lib/zluda_ptx_impl.cpp
zluda-violet: That value is the max lane ID for a specific mode – 31 for idx, bfly, and down, and 0 for up. The max lane id of the current sub-segment is calculated by replacing the bits of the lane ID not in the section mask with the bits from this value.

I don't think it makes sense for this value to be anything other than 0 and 31, and that's the values that the CUDA `__shfl_sync` uses. But it is possible for handwritten PTX to have different values, and perhaps there's a use for that, so I've replicated the CUDA logic more exactly.

ptx/lib/zluda_ptx_impl.cpp
vosen: Yes, sorry, I've started writing a comment and then started reading the code and never went back to edit it

zluda-violet:
I've checked the linking of anonymous structs for ROCm 6.4, but not 6.3 yet.

vosen (approved):

PR #421, opened 2025-07-18
Implement `nanosleep.u32`

zluda-violet:
This uses the `s_sleep` instruction to implement `nanosleep.u32`. It adds a mechanism for accessing device attributes at compile time and using them from the C++ implementation. This mechanism is used for using the clock rate in the implementation.


vosen (changes requested):
I think the approach is good, but I have two concerns:
* I think it's be more convenient to pass `CLOCK_RATE` in a separate module. Since this information is very hardware specific we could use a generic mechanism: have a module an additional set (it's so simple that it can even be a Vec<(string, usize)>) that carries that information (right now just this single entry). LLVM emission pass would emit two modules main with all the code and supporting with those hw values for ptx_impl. We have to link with ptx_impl later anyway, one more module to link does not make a difference
* Did you check if `clockRate` property is stable? We will want to cache the compiled binaries

zluda-violet:
It seems to be stable on the device I'm testing on, (it's the same today as when I checked yesterday, and each time I check), but I don't know if it might change.

zluda-violet:
@vosen I've refactored the attributes to be linked as a separate module.

ptx/src/pass/llvm/attributes.rs
```diff
@@ -0,0 +1,35 @@
+use std::ffi::CStr;
+
+use super::*;
+use super::super::*;
+use llvm_zluda::{core::*};
+
+pub(crate) fn run(attributes: Attributes) -> Result<llvm::Module, TranslateError> {
+    let context = Context::new();
```
vosen: Object lifetime here is suspicious. In general, for LLVM the Context lifetime must be longer than Module lifetime. We had problems with LLVM emission crashing because of this exact issue before.
The solution here is to have both modules share the context and keep them all together in returned object.

zluda-violet:
Hm, it doesn't seem to be being fully inlined. Any suggestions on how to fix that?

vosen:
Hmm, not sure. There's definitely a combination of attributes that gives us the outcome we want. Start with making sure that consumer (.cpp) does not declare global with `externally_initialized`

ptx/src/pass/llvm/attributes.rs
zluda-violet: The `llvm::Module` objects still own their own contexts, so I assume this code works as is but it uses two contexts. I'll refactor to use only one context, though.

zluda-violet:
Using an `extern const` from C++ where it's used and `constant` from the LLVM IR where it's defined fixes it so that it's inlined at link time.

vosen (approved):

PR #427, opened 2025-07-22
Add support for `cp.async`

zluda-violet:
Adds support for

* `cp.async`
* `cp.async.commit_group`
* `cp.async.wait_group`
* `cp.async.wait_all`

Asynchronous copy operations are only supported by AMD Instinct GPUs, so for now we lower them as synchronous copy operations. Because of this, `cp.async.commit_group`, `cp.async.wait_group`, and `cp.async.wait_all` are no-op.

### DIFF
--- a/ptx/src/pass/llvm/emit.rs
+++ b/ptx/src/pass/llvm/emit.rs
@@ -517,6 +517,8 @@ impl<'a> MethodEmitContext<'a> {
             ast::Instruction::Tanh { data, arguments } => self.emit_tanh(data, arguments),
             ast::Instruction::CpAsync { .. } => todo!(),
             ast::Instruction::CpAsyncCommitGroup { .. } => todo!(),
+            ast::Instruction::CpAsyncWaitGroup { .. } => todo!(),
+            ast::Instruction::CpAsyncWaitAll { .. } => todo!(),
             // replaced by a function call
             ast::Instruction::Bfe { .. }
             | ast::Instruction::Bar { .. }
--- a/ptx/src/pass/llvm/emit.rs
+++ b/ptx/src/pass/llvm/emit.rs
@@ -2565,29 +2565,29 @@ impl<'a> MethodEmitContext<'a> {
         let cp_size = data.cp_size;
         let src_size = data.src_size.unwrap_or(cp_size.as_u64());
 
-        let type_ = match cp_size {
+        let from_type = unsafe { LLVMIntTypeInContext(self.context, (src_size as u32) * 8) };
+
+        let to_type = match cp_size {
             ptx_parser::CpAsyncCpSize::Bytes4 => unsafe { LLVMInt32TypeInContext(self.context) },
             ptx_parser::CpAsyncCpSize::Bytes8 => unsafe { LLVMInt64TypeInContext(self.context) },
             ptx_parser::CpAsyncCpSize::Bytes16 => unsafe { LLVMInt128TypeInContext(self.context) },
         };
 
-        let load = unsafe { LLVMBuildLoad2(self.builder, type_, from, LLVM_UNNAMED.as_ptr()) };
+        let load = unsafe { LLVMBuildLoad2(self.builder, from_type, from, LLVM_UNNAMED.as_ptr()) };
+        unsafe {
+            LLVMSetAlignment(load, (cp_size.as_u64() as u32) * 8);
+        }
 
-        let masked = if src_size < cp_size.as_u64() {
-            let byte_mask = if src_size > 8 {
-                vec![u64::MAX, u64::MAX >> (64 - (src_size - 8) * 8)]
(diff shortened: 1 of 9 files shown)


ptx_parser/src/ast.rs
```diff
@@ -188,6 +188,27 @@ ptx_parser_macros::generate_instruction_type!(
                 src: T
             }
         },
+        CpAsync {
+            data: CpAsyncDetails,
+            arguments<T>: {
+                src_to: {
+                    repr: T,
+                    type: Type::Pointer(ScalarType::U32, StateSpace::Shared)
```
vosen: Change this to instead use `type: Type::Scalar(ScalarType::U32)` and `space: StateSpace::Shared` and similarly for `src_from` argument. `Type::Pointer` should be removed, any use of it has been vestigial. Remains of the old code.

ptx/src/test/ll/cp_async.ll
```diff
@@ -0,0 +1,54 @@
...
+
+1:                                                ; preds = %0
+  br label %"47"
+
+"47":                                             ; preds = %1
+  %"56" = load i64, ptr addrspace(4) %"48", align 4
+  store i64 %"56", ptr addrspace(5) %"50", align 4
+  %"57" = load i64, ptr addrspace(4) %"49", align 4
+  store i64 %"57", ptr addrspace(5) %"51", align 4
+  %2 = load i128, ptr addrspace(1) @from, align 4
```
vosen: This does not apply directly to this code, but I've realized we don't emit correct alignment on loads and stores. Please replace our uses of `LLVMBuildLoad2`/`LLVMBuildStore` with something that explicitly takes alignment argument. `LLVMSetAlignment` might work, but we might need a custom thing, see `LLVMZludaBuildAlloca` and similar for details

ptx/src/pass/llvm/emit.rs
```diff
@@ -2448,6 +2452,44 @@ impl<'a> MethodEmitContext<'a> {
...
+
+        let type_ = match cp_size {
+            ptx_parser::CpAsyncCpSize::Bytes4 => unsafe { LLVMInt32TypeInContext(self.context) },
+            ptx_parser::CpAsyncCpSize::Bytes8 => unsafe { LLVMInt64TypeInContext(self.context) },
+            ptx_parser::CpAsyncCpSize::Bytes16 => unsafe { LLVMInt128TypeInContext(self.context) },
+        };
+
+        let load = unsafe { LLVMBuildLoad2(self.builder, type_, from, LLVM_UNNAMED.as_ptr()) };
+
+        let masked = if src_size < cp_size.as_u64() {
```
vosen: Hmm wouldn't it be clearer to instead load an array of bytes with explicit alignment of `cp_size`, cast it to integer type (8 * `src_size` bits), `zext` to `cp_size` and store?

vosen (changes requested):

ptx_parser/src/ast.rs
zluda-violet: Done.

ptx/src/test/ll/cp_async.ll
zluda-violet: Sure, I've sent in a PR for this (#429) and also updated the load and store in this PR.

ptx/src/pass/llvm/emit.rs
zluda-violet: Yes, I didn't realize that LLVM ints could be an arbitrary bit width. I don't think a cast is necessary though; we can still just read a `src_size`-wide integer with alignment `cp_size`.

ptx/src/pass/llvm/emit.rs
```diff
@@ -2448,6 +2452,44 @@ impl<'a> MethodEmitContext<'a> {
...
+            ptx_parser::CpAsyncCpSize::Bytes8 => unsafe { LLVMInt64TypeInContext(self.context) },
+            ptx_parser::CpAsyncCpSize::Bytes16 => unsafe { LLVMInt128TypeInContext(self.context) },
+        };
+
+        let load = unsafe { LLVMBuildLoad2(self.builder, from_type, from, LLVM_UNNAMED.as_ptr()) };
+        unsafe {
+            LLVMSetAlignment(load, (cp_size.as_u64() as u32) * 8);
+        }
+
+        let extended = if from_type != to_type {
```
vosen: You can use `LLVMBuildZExtOrBitCast` to make it even more straightforward

zluda-violet: I don't think that's needed in this case, as `from_type` and `to_type` will always be integer types. `LLVMBuildZExt` does check for type equality though, so I can remove the check here.

vosen: Good, that's even more compact :)

PR #429, opened 2025-07-23
Emit correct alignment for loads and stores

zluda-violet:


vosen:
Nice, I thought `LLVMSetAlignment` does not work with those

Issue #424, opened 2025-07-18
My build is failing to build


Asriel115:
after successfully git cloning the program and moving into the directory building the program with "cargo xtask --release" I am now getting the follow error.

<img width="952" height="225" alt="Image" src="https://github.com/user-attachments/assets/b0c11ff9-1d67-47cd-866b-5d8b36574faf" />

vosen:
CMake failed when building LLVM. Most likely either LLVM submodule is not
checked out or there's something wrong with C++ build environment (no
make/something wrong with ninja, cmake couldn't find a C++ compiler, etc)

czw., 17 lip 2025, 22:23 użytkownik Azukuni ***@***.***>
napisał:

Asriel115:
> CMake failed when building LLVM. Most likely either LLVM submodule is not
> checked out or there's something wrong with C++ build environment (no
> make/something wrong with ninja, cmake couldn't find a C++ compiler, etc)
> 
> czw., 17 lip 2025, 22:23 użytkownik Azukuni ***@***.***>
> napisał:
> […](#)

Thanks for promptly getting back to me on this. I also tried to running "git submodule init" before building it and I'm still getting the same errors on the build process. I am unsure what the issue are in the first place, I have all the needed dependencies installed and followed the build instructions to the "T". I am not sure what information I would need to provide for a build error like this. 

<img width="898" height="220" alt="Image" src="https://github.com/user-attachments/assets/ff52b78a-45cb-48b6-80db-abbd64a06edf" />

vosen:
It's weird that you are not getting any cmake log spew. Do you have cmake installed? Does `cmake --version` print something?
```
vosen@home:~/dev/zluda/ptx/lib$ cmake --version
cmake version 4.0.3

CMake suite maintained and supported by Kitware (kitware.com/cmake).
```

Also try running with env var `RUST_BACKTRACE=full` and building without `--release`: just `cargo xtask`

Asriel115:
<img width="716" height="156" alt="Image" src="https://github.com/user-attachments/assets/f21acf90-069b-44fa-813b-b8d0de123075" />

I've had CMake installed this whole time. I made sure all my dependencies were installed before I submitted this issue. Though I am getting a different error now that I tried your suggestion.

````thread 'main' panicked at /home/azuma/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cmake-0.1.51/src/lib.rs:1100:5:

  command did not execute successfully, got: exit status: 1

[...]

PR #432, opened 2025-07-24
Add issue form for zluda_trace logs

zluda-violet:


.github/ISSUE_TEMPLATE/zluda_dump.yml
```diff
@@ -0,0 +1,56 @@
...
+      description: What operating system are you using? (e.g., distribution and version)
+      placeholder:  "example: Ubuntu 22.04.5 LTS"
+    validations:
+      required: true
+  - type: input
+    id: gpu
+    attributes:
+      label: GPU
+      description: What GPU are you using?
+      placeholder:  "example: NVIDIA GeForce RTX 5080"
```
vosen: A better example would be "AMD Radeon RX 6600" :)

.github/ISSUE_TEMPLATE/zluda_dump.yml
```diff
@@ -0,0 +1,56 @@
...
+  - type: textarea
+    id: logs
+    attributes:
+      label: zluda_dump logs (tarball/zip file)
+      description: |
+        Please create a tarball (`.tar.gz`) or zip file (`.zip`) of your log directory and attach
+        it here. You can drag and drop files directly into the comment box.
+      placeholder: Attach file (e.g., drag and drop)
+    validations:
+      required: true
```
vosen: It's better to leave it optional to account for people that come with build errors

vosen (changes requested):

.github/ISSUE_TEMPLATE/zluda_dump.yml
zluda-violet: Build errors for ZLUDA or zluda_dump? We can create a separate template or form for that (and people can still create issues with no template – that's what `blank_issues_enabled: true` in config.yml does)

.github/ISSUE_TEMPLATE/zluda_dump.yml
zluda-violet: Are we assuming people are using zluda_dump with CUDA to trace all library calls that will be made, or with ZLUDA to trace the ones that are currently implemented?

vosen: Most of our users will have just an AMD GPU, but we should ask them to produce the log from CUDA GPU if they have access to one

.github/ISSUE_TEMPLATE/zluda_dump.yml
vosen: Build errors for ZLUDA, e.g. #424
I think another template would be too much. One template for the most typical "my app does not work, please help" is enough. And a way to enter any issue that does not fit this mold (like build errors)

vosen (approved):

PR #436, opened 2025-07-25
Refactor initialized state mechanism into a common crate

zluda-violet:
We will likely want to use this in zluda_blas and the other performance libraries


zluda-violet:
I'm not actually sure if this is necessary.

zluda-violet:
Though there might be some other things that benefit from being refactored into a common crate.

PR #440, opened 2025-07-28
Add support for `hipblas` to `zluda_bindgen`

zluda-violet:
One step of several for adding cublas support


vosen:
Windows should be linkable without having the .dll or .lib present. Look into `raw-dylib` linking

ext/rocblas-sys/build.rs
```diff
@@ -0,0 +1,19 @@
...
+        // let env = env::var("CARGO_CFG_TARGET_ENV")?;
+        // if env == "msvc" {
+        //     let mut path = PathBuf::from(env::var("CARGO_MANIFEST_DIR")?);
+        //     path.push("lib");
+        //     println!("cargo:rustc-link-search=native={}", path.display());
+        // } else {
+        //     println!("cargo:rustc-link-search=native=C:\\Windows\\System32");
+        // };
+    } else {
+        println!("cargo:rustc-link-lib=dylib:+verbatim=librocblas.so.4");
```
vosen: I prefer `dylib=rocblas`. There's no real reason to introduce breaking changes into rocblas, so i think it's fine to link to non-versioned library. Unless there's some other reason?

zluda-violet: Nope, I was just copying https://github.com/vosen/ZLUDA/blob/adc4673/ext/hip_runtime-sys/build.rs#L16

zluda-violet: Done.

vosen: Ah, hipruntime is special. There are can be two versions present in the system with different ABI, so in that case version-specific linking is warranted

vosen (changes requested):
This code redefines `hipStream_t` and `hipEvent_t`. It should instead use existing types from `hip_runtime-sys`, otherwise you are setting yourself up for trouble down the line e.g. when  you extract `hip_runtime-sys::hipStream_t` from CUstream you will need to cast it to `rcoblas-sys::hipStream_t`

zluda_bindgen/src/main.rs
```diff
@@ -921,6 +922,60 @@ fn generate_hip_runtime(output: &PathBuf, path: &[&str]) {
...
+    module.items = module
+        .items
+        .into_iter()
+        .filter_map(|item| match item {
+            Item::Const(const_) => converter.get_const(const_).map(Item::Const),
+            Item::Use(use_) => converter.get_use(use_).map(Item::Use),
+            Item::Type(type_) => converter.get_type(type_).map(Item::Type),
+            Item::ForeignMod(mut extern_) => {
+                extern_.attrs.push(
+                    parse_quote!(#[cfg_attr(windows, link = "rocblas_4", kind = "raw-dylib")]),
```
vosen: This should be "rocblas" instead. On Windows ROCm appends version number only to amdhip, comgr and hiprtc.

zluda_bindgen/src/main.rs
```diff
@@ -921,6 +922,60 @@ fn generate_hip_runtime(output: &PathBuf, path: &[&str]) {
...
+fn generate_rocblas(output: &PathBuf, path: &[&str]) {
+    let rocblas_header = new_builder()
+        .header("/opt/rocm/include/rocblas/rocblas.h")
+        .allowlist_type("^rocblas.*")
+        .allowlist_function("^rocblas.*")
+        .allowlist_var("^rocblas.*")
+        .must_use_type("rocblas_status")
+        .constified_enum("rocblas_status_")
+        .new_type_alias("^rocblas_handle$")
+        .new_type_alias("^hipStream_t$")
```
vosen: Is this line and the one below still necessary?

zluda_bindgen/src/main.rs
zluda-violet: Done.

zluda_bindgen/src/main.rs
zluda-violet: This converts the type into a struct so that `remove_type` will find it. On second thought it's probably better to remove this and update `remove_type` to handle types, so I've done that.

PR #444, opened 2025-07-28
Convert CUDA performance lib statuses to Rust result types

zluda-violet:
These changes replicate how the main library is handled. cuDNN still needs to have zluda_bindgen run and zluda_dump_common updated.


cuda_types/src/cufft.rs
```diff
@@ -50,62 +50,8 @@ pub struct cudaLibXtDesc_t {
...
-}
-impl cufftResult_t {
-    pub const CUFFT_NOT_SUPPORTED: cufftResult_t = cufftResult_t(16);
-}
-#[repr(transparent)]
 #[must_use]
-#[derive(Debug, Copy, Clone, Hash, PartialEq, Eq)]
-pub struct cufftResult_t(pub ::core::ffi::c_uint);
-pub use self::cufftResult_t as cufftResult;
+pub type cufftResult_t = ::core::ffi::c_uint;
```
vosen: There's something wrong here, this type shouldn't be an alias to `c_uint`

vosen (changes requested):

cuda_types/src/cufft.rs
zluda-violet: This is the same behavior that cuda.rs currently has for `underlying_type`: https://github.com/vosen/ZLUDA/blob/master/cuda_types/src/cuda.rs#L4338. I'll make `get_type` ignore `underlying_type`.

Note that `cufftResult` and `cufftResult_t` are the typedef and enum, respectively, which is the reverse of the convention from some of the other libraries

zluda-violet: Done.

cuda_types/Cargo.toml
```diff
@@ -8,3 +8,4 @@ edition = "2021"
 cuda_macros = { path = "../cuda_macros" }
 hip_runtime-sys = { path = "../ext/hip_runtime-sys" }
 bitflags = "2.9.1"
+rocblas-sys = { version = "0.1.0", path = "../ext/rocblas-sys" }
```
vosen: Minor, but is there a reason to have version present? I can't think of a scenario where it would be helpful

zluda-violet: No, `cargo add` added that automatically.

vosen (approved):

PR #447, opened 2025-07-29
Force loading ZLUDA through LD_PRELOAD

vosen:
Some applications (ok, literally just pytorch) decided that it's a great idea to distribute whole CUDA driver and link it with `DT_RPATH`. This defeats the usual `LD_LIBRARY_PATH` mechanism.
This code defeats that evil mechanism through any means necessary


zluda_preload/README.md
```diff
@@ -0,0 +1,21 @@
+This crate is a last resort Linux-specific solution.
+Most of the time we can inject ourselves into a process by having users
+set `LD_LIBRARY_PATH`.
+Unfortunately, there is software out there which dynamically links to CUDA and
+and CUDA performance libraries using RPATH. On Linux, dynamic linker operates
```
zluda-violet: nit: redundant and

zluda_preload/src/lib.rs
```diff
@@ -0,0 +1,128 @@
+use std::{
+    ffi::{c_char, c_int, c_void, CStr},
+    mem,
+    path::PathBuf,
+    ptr::{self, NonNull},
+    sync::LazyLock,
+};
+use unwrap_or::unwrap_some_or;
+
+const RTLD_NEXT: *mut c_void = -1isize as _;
```
zluda-violet, line 12: Is there a reason you're not doing this?
```suggestion
const RTLD_NEXT: *mut c_void = usize::MAX as _;
```

zluda-violet: Never mind, I see this is how it's defined in glibc. https://github.com/lattera/glibc/blob/895ef79e04a953cac1493863bcae29ad85657ee1/dlfcn/dlfcn.h#L35

Maybe add a comment explaining that?

zluda_preload/src/lib.rs
```diff
@@ -0,0 +1,128 @@
...
+    "libnvidia-ml.so.1",
+];
+
+static GLOBALS: LazyLock<(
+    Option<unsafe extern "C" fn(*const c_char, c_int) -> DlopenResult>,
+    Option<[Vec<u8>; FILES_FOR_REDIRECT.len()]>,
+)> = LazyLock::new(|| {
+    let dlopen_next = unsafe { mem::transmute(dlsym(RTLD_NEXT, c"dlopen".as_ptr())) };
+    let mut self_dlinfo = unsafe { mem::zeroed::<DLInfo>() };
+    let self_dir = if unsafe { dladdr(dlopen as _, &mut self_dlinfo) } != 0 {
```
zluda-violet: Maybe add a comment explaining what this is: a list of paths to the ZLUDA shared objects we're redirecting. It might be more readable if `GLOBALS` held a struct instead of a tuple, with comments explaining each field:

```rust
struct Globals {
    /// The original `dlopen` implementation from libdl.
    dlopen_next: Option<unsafe extern "C" fn(*const c_char, c_int) -> DlopenResult>,
    /// The full paths to the shared objects we are redirecting from `FILES_FOR_REDIRECT`.
    replacement_paths: Option<[Vec<u8>; FILES_FOR_REDIRECT.len()]>,
}
```

zluda_preload/src/lib.rs
```diff
@@ -0,0 +1,128 @@
...
+pub const RTLD_GLOBAL: c_int = 0x100;
+pub const RTLD_LAZY: c_int = 1;
+
+#[ctor::ctor]
+unsafe fn ctor() {
+    let (dlopen_next, replacement_paths) = &*GLOBALS;
+    let dlopen_next = unwrap_some_or!(dlopen_next, return);
+    let replacement_paths = unwrap_some_or!(replacement_paths, return);
+    for replacement in replacement_paths.into_iter() {
+        dlopen_next(replacement.as_ptr().cast(), RTLD_GLOBAL | RTLD_LAZY).ok();
```
zluda-violet, line 97: Could you add some explanation of why we do this preloading in addition to redirecting `dlopen`?

zluda_trace/src/os_unix.rs
```diff
@@ -1,12 +1,19 @@
...
 pub(crate) const LIBCUDA_DEFAULT_PATH: &str = "/usr/lib/x86_64-linux-gnu/libcuda.so.1";
 
 pub unsafe fn load_library(libcuda_path: &str) -> *mut c_void {
+    let zluda_dlopen_noredirect =
+        libc::dlsym(libc::RTLD_DEFAULT, c"zluda_dlopen_noredirect".as_ptr());
+    let zluda_dlopen_noredirect = mem::transmute::<
+        _,
+        Option<unsafe extern "C" fn(*const c_char, c_int) -> *mut c_void>,
+    >(zluda_dlopen_noredirect);
+    let dlopen = zluda_dlopen_noredirect.unwrap_or(libc::dlopen);
```
zluda-violet: Is there a reason this isn't using `zluda_trace_common::dlopen_local_noredirect`?

zluda-violet (commented):
This generally makes sense to me. I've left a couple comments.

zluda_preload/src/lib.rs
vosen: Done

zluda_preload/src/lib.rs
vosen: Done

zluda_trace/src/os_unix.rs
vosen: No reason, `dlopen_local_noredirect` covers Windows too. I've changed to use it

zluda_preload/README.md
vosen: Done

zluda-violet (approved):
LGTM!

PR #451, opened 2025-07-30
Check Rust formatting on pull requests

zluda-violet:
This should help us maintain consistent style, without having unrelated style changes in pull requests from running `rustfmt`.


zluda-violet:
This doesn't seem to have a way to exclude specific paths :/
I'll probably close this for now.

zluda-violet:
Disabling the directories with generated code using `.rustfmt.toml` should work.

PR #452, opened 2025-07-30
Move `FromCuda` and `ZludaObject` into a common crate

zluda-violet:
It would be convenient to use `FromCuda` and `ZludaObject` from the convenience libraries. I've refactored them to use a generic error type, and moved them into a common trait.


zluda_common/src/lib.rs
```diff
@@ -0,0 +1,278 @@
...
+            let result = unsafe { self.data.assume_init_mut().drop_checked() };
+            unsafe { MaybeUninit::assume_init_drop(&mut self.data) };
+            Ok(result)
+        } else {
+            Err(T::LIVENESS_FAIL)
+        }
+    }
+}
+
+/// Cast arbitrary memory to a [`LiveCheck`] value.
```
vosen: Not really arbitrary memory, but casts a reference to unsafe handle to a reference to an underlying object while preserving the lifetime. Preserving the lifetime is the key here, `mem::transmute` of a reference will infer the lifetime, very likely giving you a 'static lifetime

zluda-violet: Ah right, I was thinking it's arbitrary in the sense that the C++ code calling `cuModuleGetFunction` can pass in whatever address it wants, but that cast happens in `from_cuda_object`.

zluda-violet: I've rewritten the comment.

PR #455, opened 2025-07-30
Use `FromCuda` in `zluda_blas`

zluda-violet:
Not all cublas calls return cublasStatus_t, and `?` cannot be used for those. There's a couple of possibilities for handling this, but for now I'm just converting the arguments to `Option` and letting the handler deal with them.

### DIFF
--- a/zluda_blas/src/lib.rs
+++ b/zluda_blas/src/lib.rs
@@ -1,7 +1,5 @@
 mod r#impl;
 
-use cuda_types::cublas::cublasError_t;
-
 macro_rules! unimplemented {
     ($($abi:literal fn $fn_name:ident( $($arg_id:ident : $arg_type:ty),* ) -> $ret_type:ty;)*) => {
         $(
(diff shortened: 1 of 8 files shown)


zluda_blas/src/lib.rs
```diff
@@ -20,15 +22,30 @@ macro_rules! implemented {
...
+}
+
+macro_rules! implemented_and_always_succeeds {
+    ($($abi:literal fn $fn_name:ident( $($arg_id:ident : $arg_type:ty),* ) -> $ret_type:ty;)*) => {
+        $(
+            #[cfg_attr(not(test), no_mangle)]
+            #[allow(improper_ctypes)]
+            #[allow(improper_ctypes_definitions)]
+            pub unsafe extern $abi fn $fn_name ( $( $arg_id : $arg_type),* ) -> $ret_type {
+                cuda_macros::cublas_normalize_fn!( crate::r#impl::$fn_name ) ($(zluda_common::FromCuda::<_, cublasError_t>::from_cuda(&$arg_id).ok()),*)
```
vosen: Why convert to Result and Option? Wouldn't it be cleaner to pass those arguments as-is, no conversion? They are ignored anyway

vosen:
Sorry about those two commits, I did not realize I have your branch checked out

zluda_blas/src/lib.rs
zluda-violet: Good point. Done.

PR #457, opened 2025-07-31
Implement cublas functions needed for llm.c

zluda-violet:
Implements:

* `cublasCreate_v2`
* `cublasSetMathMode`
* `cublasSgemmStridedBatched`
* `cublasSgemm_v2`
* `cublasDestroy_v2`

These are the final API calls needed for llm.c's `test_gpt2fp32cu`.

### DIFF
--- a/zluda_blas/src/impl.rs
+++ b/zluda_blas/src/impl.rs
@@ -63,8 +63,8 @@ pub(crate) fn get_cudart_version() -> usize {
     todo!()
 }
 
-pub(crate) fn set_math_mode(_handle: &Handle, _mode: cublasMath_t) -> cublasStatus_t {
-    // TODO: hipblas implements this but rocblas does not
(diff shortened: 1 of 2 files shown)


zluda_blas/src/impl.rs
```diff
@@ -25,3 +62,94 @@ pub(crate) fn xerbla(_sr_name: *const ::core::ffi::c_char, _info: ::core::ffi::c
 pub(crate) fn get_cudart_version() -> usize {
     todo!()
 }
+
+pub(crate) fn set_math_mode(_handle: &Handle, _mode: cublasMath_t) -> cublasStatus_t {
+    // TODO: hipblas implements this but rocblas does not
```
vosen: But rocblas does implement it? I can see `rocblas_set_math_mode` in bindings

zluda-violet: Oh woops, I looked in the documentation but not the bindings

zluda-violet: Fixed.

PR #463, opened 2025-08-02
Implement `shf` instruction

zluda-violet:
`llvm.fsh*.i32` behaves the same as `shf.*.wrap.b32`. `shf.*.clamp.b32` must be implemented by checking for the shift amount being greater than or equal to 32, and returning the maximally shifted 32 bits in that case.


vosen (approved):

vosen:
Approved, just reformat one file

PR #465, opened 2025-08-05
Implement kernel cache

### LINKED ISSUE #336: Kernel cache
Expected outcome:
After PTX module have been compiled once, it gets cached on disk in an SQLite database. Before kernel gets compiled we look in cache to check if the kernel has been previously compiled

Comments:
* Key should be hash (BLAKE3) of module text + compiler version + ZLUDA version + device (gfxXXXX) + flags (debug/release, windows/linux, compiler switches)
* We will want to eventually support mechanism similar to `CUDA_​CACHE_​MAXSIZE`. This does not have to be implemented yet, but db should at least contain necessary information for evicting cache: time of last use for each entry and total size of all the kernels
* You can get compile version from comgr by running a preprocessor on a file containing preprocessor directive that resolves to full clang version. It contains version and ROCm LLVM hash
* ZLUDA version should be current git commit hash. There are several crates for it, last time I used vergen

vosen:
Will close #336 when merged

### DIFF
--- a/ptx/src/test/spirv_run/mod.rs
+++ b/ptx/src/test/spirv_run/mod.rs
@@ -643,7 +643,9 @@ fn run_hip<Input: From<u8> + Copy + Debug, Output: From<u8> + Copy + Debug + Def
         unsafe { hipGetDevicePropertiesR0600(&mut dev_props, dev) }.unwrap();
         let elf_module = comgr::compile_bitcode(
             &comgr,
-            unsafe { CStr::from_ptr(dev_props.gcnArchName.as_ptr()) }.to_str().unwrap(),
+            unsafe { CStr::from_ptr(dev_props.gcnArchName.as_ptr()) }
+                .to_str()
+                .unwrap(),
             &*module.llvm_ir.write_bitcode_to_memory(),
             &*module.attributes_ir.write_bitcode_to_memory(),
             module.linked_bitcode(),
--- a/zluda_cache/src/schema.rs
+++ b/zluda_cache/src/schema.rs
@@ -20,7 +20,4 @@ diesel::table! {
     }
 }
 
-diesel::allow_tables_to_appear_in_same_query!(
-    globals,
-    modules,
-);
+diesel::allow_tables_to_appear_in_same_query!(globals, modules,);


vosen:
The code does not cover cache eviction. When a cache gets too large (based on `CUDA_CACHE_MAXSIZE`) we should remove entries not used recently. It can wait for now

zluda/src/impl/module.rs
```diff
@@ -65,26 +65,95 @@ fn get_ptx(image: *const ::core::ffi::c_void) -> Result<String, CUerror> {
...
+    let hip_properties = get_hip_properties()?;
+    let gcn_arch = get_gcn_arch(&hip_properties)?;
     let attributes = ptx::Attributes {
-        clock_rate: props.clockRate as u32,
+        clock_rate: hip_properties.clockRate as u32,
     };
+    let mut cache_with_key = global_state.cache_path.as_ref().and_then(|p| {
+        let cache = zluda_cache::ModuleCache::open(p)?;
+        let key = get_cache_key(global_state, gcn_arch, &text, &attributes)?;
+        Some((cache, key))
```
zluda-violet, line 76: Maybe just call `cache.get_binary(key)` here and get rid of `load_cached_binary`?

zluda/src/impl/driver.rs
```diff
@@ -52,8 +54,11 @@ pub(crate) fn global_state() -> Result<&'static GlobalState, CUerror> {
             let mut device_count = 0;
             unsafe { hipGetDeviceCount(&mut device_count) }?;
             let comgr = Comgr::new().map_err(|_| CUerror::UNKNOWN)?;
+            let comgr_clang_version =
+                comgr::get_clang_version(&comgr).map_err(|_| CUerror::UNKNOWN)?;
```
zluda-violet, line 58: Why are we using Clang version, specifically, considering that we're not actually using Clang for compiling PTX? Is it just easier to check?

vosen: Yes, I don't know of a good way to extract LLVM version, but we can extract clang version with builtin preprocessor define:
```
echo | /opt/rocm/llvm/bin/clang -dM -E -
...
#define __clang_version__ "19.0.0git (https://github.com/RadeonOpenCompute/llvm-project roc-6.4.0 25133 c7fe45cf4b819c5991fe208aaa96edf142730f1d)"
...
```

which conveniently contains git commit used to build comgr's Clang&LLVM

zluda/src/impl/module.rs
vosen: Good idea, but I've tried it just now and it makes code less readable.
Even if we load binary together with cache & key, we have to split binary from cache&key anyway. Maybe if we had `Option<(X, Y)> -> (Option<X>, Option<Y>)` in std

PR #466, opened 2025-08-05
More descriptive syntax errors

zluda-violet:
Writes more descriptive syntax errors. I'm generating a `Display` implementation for parser enums for use in the errors, and I've implemented it for `ParsedOperand` and `Immediate` as well. This should make it a little easier to inspect the AST state while debugging as well.

### DIFF
--- a/ptx_parser/src/ast.rs
+++ b/ptx_parser/src/ast.rs
@@ -1145,7 +1145,7 @@ where
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             ParsedOperand::Reg(id) => write!(f, "{}", id)?,
-            ParsedOperand::RegOffset(id, o) => write!(f, "[{}+{}]", id, o)?,
+            ParsedOperand::RegOffset(id, o) => write!(f, "{}+{}", id, o)?,
             ParsedOperand::Imm(imm) => write!(f, "{}", imm)?,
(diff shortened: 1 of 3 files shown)


ptx_parser/src/ast.rs
```diff
@@ -1138,6 +1138,40 @@ impl<Ident> ParsedOperand<Ident> {
...
 }
 
+impl<Ident> std::fmt::Display for ParsedOperand<Ident>
+where
+    Ident: std::fmt::Display,
+{
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            ParsedOperand::Reg(id) => write!(f, "{}", id)?,
+            ParsedOperand::RegOffset(id, o) => write!(f, "[{}+{}]", id, o)?,
```
vosen: `[` `]` feel unnecessary here, AFAIK grammar does not prevent RegOffset from showing in non-bracketized positions. Though it might might make it more readable in real world?
I'm leaving it up to you

vosen (approved):
Looks good.
Only tangentially related, but writing it down since I've been thinking about it recently: in release mode we will want to return semi-correct result even in the face of parsing errors. Specifically if there is an error within the scope of a function mark this function as invalid and continue parsing.
This is purely for release mode where we might not support a single instruction in a single function, but it might be a function that is actually unused and still enable most functionality for the user.

PR #468, opened 2025-08-06
Add support for fp8 to `cvt`

zluda-violet:
This implements specifically the fp8 conversion instructions needed by llm.c:

* `cvt.rn.satfinite{.relu}.f8x2type.f32`
* `cvt.rn{.relu}.f16x2.f8x2type`

It uses HIP's fp8 and fp16 headers: https://rocm.docs.amd.com/projects/HIP/en/docs-develop/reference/low_fp_types.html#fp8-quarter-precision.


ptx/lib/zluda_ptx_impl.cpp
```diff
@@ -335,4 +339,28 @@ extern "C"
         else
             return value;
     }
+
+    __hip_fp8x2_e4m3 FUNC(cvt_f32_to_e4m3_satfinite)(float a, float b)
```
vosen: Did you check .ll if the generated signatures match expected datatypes? Especially when using HIP types (even something innocent like float2) you can get a type that has a very different LLVM bitcode than you'd expect

ptx/src/pass/replace_instructions_with_function_calls.rs
```diff
@@ -185,6 +185,50 @@ fn run_instruction<'input>(
...
+                ptx_parser::ScalarType::E4m3x2 => "e4m3",
+                ptx_parser::ScalarType::E5m2x2 => "e5m2",
+                _ => unreachable!(),
+            };
+            // Conversions from f32 to f8 must have two source arguments.
+            // satfinite is mandatory for conversions to e4m3x2.
+            to_call(
+                resolver,
+                fn_declarations,
+                format!("cvt_f32_to_{}_satfinite", to).into(),
```
vosen: I'd like you to rename to strictly follow PTX naming convention. So instead call it `cvt_<modifier>_<dst_type>_<src_type>`. Even if it's not intuitive, it is consistent across the code base.

ptx/src/pass/llvm/mod.rs
```diff
@@ -133,12 +133,15 @@ fn get_scalar_type(context: LLVMContextRef, type_: ast::ScalarType) -> LLVMTypeR
...
-        },
+        ast::ScalarType::B16
+        | ast::ScalarType::U16
+        | ast::ScalarType::S16
+        | ast::ScalarType::E4m3x2
+        | ast::ScalarType::E5m2x2 => unsafe { LLVMInt16TypeInContext(context) },
+        ast::ScalarType::S32
+        | ast::ScalarType::B32
+        | ast::ScalarType::U32
+        | ast::ScalarType::F16x2 => unsafe { LLVMInt32TypeInContext(context) },
```
vosen: This line is definitely wrong, f16x2 should compile to two element vector of f16. This is not a problem here, but once we have something like `fma.f16x2` it will explode

vosen:
One more thing I forgot to put in the review: please add PR for those two conversions into tests here: https://github.com/vosen/ptx_tests and make sure they pass

vosen:
Also my reference for implicit conversions for `.??x2` types: https://docs.nvidia.com/cuda/parallel-thread-execution/#packed-floating-point-data-types

zluda-violet:
Sorry for the force push, it was accidental (`jj` defaults to force push and will push everything your current change is on top of)

ptx/lib/zluda_ptx_impl.cpp
zluda-violet: Yes, it's `i16` for the FP8x2 vector types.

ptx/src/pass/replace_instructions_with_function_calls.rs
zluda-violet: Done.

ptx/src/pass/llvm/mod.rs
zluda-violet: Fixed.

zluda-violet:
I've added the requested fixes.

As for refactoring the `replace_instructions_with_functions` pass, at a certain point I think it might just be simpler to implement the ABI logic from Clang.

zluda-violet:
I'm adding the appropriate tests to https://github.com/vosen/ptx_tests next.

vosen:
Merged because even though I posted comments in `ptx_tests`, my suggestions do not change tests results

PR #472, opened 2025-08-08
Add test for unrecognized statement error with vector braces

zluda-violet:
The old code using`take_till_inclusive` assumed that a right brace would be the end of a block and therefore never part of a statement. However, some PTX statements can include vector operands. This meant that any unrecognized statement with a vector operand would backtrace and eventually produce an unhelpful context error rather than an `UnrecognizedStatement` error.

This pull request also adds a mechanism for testing parser errors.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,6 +9,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
 
 [[package]]
+name = "ahash"
+version = "0.8.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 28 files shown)


ptx/src/test/parse_fail/mod.rs
```diff
@@ -0,0 +1,34 @@
...
+            #[test]
+            fn [<$fn_name _llvm>]() -> Result<(), Box<dyn std::error::Error>> {
+                let ptx = read_test_file!(concat!(stringify!($fn_name), ".ptx"));
+                test_parse_fails(&ptx, $expected_err)
+            }
+        }
+    };
+}
+
+test_ptx_parse_fails!(
```
vosen: There are error recovery tests already in `zluda/ptx_parser/src/lib.rs`. Is there a reason to move this one here?

zluda-violet: No, I just didn't realize or forgot those were there. I've removed this directory and moved the test there.

PR #477, opened 2025-08-19
Add pass test mechanism for insert_implicit_conversions

zluda-violet:
Adds the ability to test specific passes, by comparing a text representation of our PTX IR. Because we don't have the ability to parse into any pass-specific IR, this starts from PTX and runs all the passes until the desired pass under test.

The AST description is extended to include a `display` expression that handles displaying the node. The printer will fall back to printing something like `<Abs> %1, %2` if this field is not present. Since these are only being used for tests we can add them as-needed.

### DIFF
--- a/ptx/src/pass/mod.rs
+++ b/ptx/src/pass/mod.rs
@@ -29,6 +29,9 @@ mod replace_instructions_with_functions_fp_required;
 mod replace_known_functions;
 mod resolve_function_pointers;
 
+#[cfg(test)]
+mod test;
+
 static ZLUDA_PTX_IMPL: &'static [u8] = include_bytes!("../../lib/zluda_ptx_impl.bc");
 const ZLUDA_PTX_PREFIX: &'static str = "__zluda_ptx_impl_";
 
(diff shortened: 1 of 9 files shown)


ptx/src/pass/test/insert_implicit_conversions/default.ptx
```diff
@@ -0,0 +1,22 @@
...
+
+.func (.reg .b32 output) default (
+    .reg .u32 input
+)
+{
+    mov.b32 output, input;
+    ret;
+}
+
+// output:
```
vosen: We all trained ourselves to ignore comments when skimming source code. Maybe use something more distinctive as the separator, like `%%%` or `@@@` or `###` or `&&&`?

ptx/src/pass/mod.rs
```diff
@@ -587,6 +624,19 @@ enum ConversionKind {
     AddressOf,
 }
 
+impl std::fmt::Display for ConversionKind {
```
vosen: You can derive this with strum:
```
#[derive(strum_macros::Display)]
#[strum(serialize_all = "snake_case")]
```

vosen (changes requested):
In this test I want formatting to output
```
.reg.u32 %4 = zluda.convert_implicit.default.reg.b32.reg.u32 %3;
```
I think it can be done by:
* Instead of implementing Display for `DisplayDirective2Vec`, `Directive2`,  `Statement`, etc. use plain functions that also take resolver
* Keep passing this resolved down to the Statement. Statement can be formatted by visiting the arguments (we already have `.visit(...)`) and first printing all destination arguments (it's an argument to `.visit(...)` that are unnamed (look it up in the resolver). Then printing instruction itself (without arguments) and then printing all other arguments
* I am not sure how to allow for an override in case arguments need special formatting, but that is something to worry about some other time (at least brackets in `ld` and `st` can be inferred from the state space)

Let me know what you think

ptx/src/pass/mod.rs
zluda-violet: Done.

ptx/src/pass/test/insert_implicit_conversions/default.ptx
zluda-violet: Done.

zluda-violet:
Your approach sounds like a good idea, and I've implemented it. I would like this mechanism to be general enough to print IR for every pass, for two reasons: being able to print a more human-readable representation of instructions in the debugger, and being able to implement something like LLVM's `-print-after-all` at some point. Most of the PTX passes are fairly simple though so that might not be necessary.

For now I've made this more specific, to work on statements and instructions with `SpirvWord` operands.

vosen (approved):

vosen:
One more reason to increase the printing coverage would be for 32 bit PhysX on RX 5000 and newer.
The compilation process for that is to parse, run pass for 32bit-on-64bit PTX (change some instructions, insert implicit arguments) and then convert back to PTX text (using the printing)

PR #483, opened 2025-08-27
zoc: Stepwise compilation and output

JoelleJS:
Splits `CompileSourceToExecutable` into `CompileSourceWithDeviceLibsToBc`, `CodegenBcToRelocatable` and `LinkRelocatableToExecutable` when ZOC is used and outputs the intermediate artifacts:
- `CompileSourceWithDeviceLibsToBc` outputs `<name>.with-device-libs.ll`
- `CodegenBcToRelocatable` outputs `<name>.relocatable.elf`
- Relocatable is also disassembled, outputs `<name>.relocatable.asm`
- `LinkRelocatableToExecutable` outputs `<name>.elf` (unchanged)

### DIFF
--- a/comgr/src/lib.rs
+++ b/comgr/src/lib.rs
@@ -259,30 +259,74 @@ pub fn compile_bitcode(
             c"-inlinehint-threshold=3250",
         ]
     };
-    let exec_data_set: DataSet;
+    let options = common_options.chain(opt_options);
     let executable: Result<Vec<u8>, Error>;
     if let Some(hook) = compiler_hook {
-        // Run compiler hook for executable
+        // CompileSourceWithDeviceLibsToBc
+        let action_info = ActionInfo::new(comgr)?;
+        action_info.set_isa_name(gcn_arch)?;
+        action_info.set_language(Language::LlvmIr)?;
+        action_info.set_options(options.clone())?;
+        let dataset = comgr.do_action(
+            ActionKind::CompileSourceWithDeviceLibsToBc,
+            &action_info,
+            &linked_data_set,
+        )?;
+        let data = dataset.get_content(comgr, DataKind::Bc, 0)?;
+        let data = ptx::bitcode_to_ir(data);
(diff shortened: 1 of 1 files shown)


comgr/src/lib.rs
```diff
@@ -249,32 +259,78 @@ pub fn compile_bitcode(
...
+        action_info.set_isa_name(gcn_arch)?;
+        action_info.set_language(Language::LlvmIr)?;
+        action_info.set_options(options.clone())?;
+        let dataset = comgr.do_action(
+            ActionKind::CompileSourceWithDeviceLibsToBc,
+            &action_info,
+            &linked_data_set,
+        )?;
+        let data = dataset.get_content(comgr, DataKind::Bc, 0)?;
+        let data = ptx::bitcode_to_ir(data);
```
vosen, line 276: This fails with `Invalid attribute group entry (Producer: 'LLVM19.0.0git' Reader: 'LLVM 17.0.6')`, which makes sense. LLVM that ZLUDA links to is (deliberately) 17.0.6, while comgr is whatever ships with HIP (18 or 19). Older LLVM can't read newer bitcode.
Not sure if there is a  good solution to this. This should be handled by comgr itself and it does not look like comgr exposes this capability? There's a promising `AMD_COMGR_ACTION_DISASSEMBLE_BYTES_TO_SOURCE`, but it seems to be used only for input filtering, disassembly seems to be an equivalent of `objdump`, so ELF/COFF only.

Not sure if there's a good solution is here. On linux one could try to to sniff path to `llvm-dis`? Start with the comgr path via `dlinfo` and try to call llvm-dis relative to it (on my machine comgr is in `/opt/rocm/lib` and llvm-dis is in `/opt/rocm/llvm/bin/`

If there is no good solution feel free to just comment it out with appropriate note

vosen (changes requested):

vosen:
Closing because the master branch moved too far and in the meantime zoc now has a way to get the assembly (though not as fine-grained as in the PR)

PR #488, opened 2025-08-29
Support immediates in vector operands

zluda-violet:
I think this might not have been what we discussed, so let me know if I should take a different approach.

### DIFF
--- a/ptx/src/pass/expand_operands.rs
+++ b/ptx/src/pass/expand_operands.rs
@@ -237,7 +237,9 @@ impl<'a, 'input> FlattenArguments<'a, 'input> {
     ) -> Result<SpirvWord, TranslateError> {
         let (width, scalar_t, state_space) = match type_space {
             Some((ast::Type::Vector(width, scalar_t), space)) => (*width, *scalar_t, space),
-            Some((ast::Type::Scalar(scalar_t), space)) => {
+            Some((ast::Type::Scalar(scalar_t), space))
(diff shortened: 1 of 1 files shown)


ptx/src/pass/expand_operands.rs
```diff
@@ -230,15 +230,31 @@ impl<'a, 'input> FlattenArguments<'a, 'input> {
...
         &mut self,
-        vector_elements: Vec<SpirvWord>,
+        vector_elements: Vec<ast::RegOrImmediate<SpirvWord>>,
         type_space: Option<(&ast::Type, ast::StateSpace)>,
         is_dst: bool,
         relaxed_type_check: bool,
     ) -> Result<SpirvWord, TranslateError> {
         let (width, scalar_t, state_space) = match type_space {
             Some((ast::Type::Vector(width, scalar_t), space)) => (*width, *scalar_t, space),
+            Some((ast::Type::Scalar(scalar_t), space)) => {
```
vosen: From my experiments this should be only valid if `scalar_t` is a binary type (`.b32`, `.b64`, etc), please add a check.

BTW, this is still slightly different logic from NVIDIA compiler because they will error on
```
	.reg .b64 	    in_addr;
    .reg .u32       tempu;
    .reg .f32       tempf;
    mov.b64    in_addr, {tempu, tempf};
```
but their logic is stupid, so let's go with this code

vosen (changes requested):
Minor addition required and please resolve the merge conflict

PR #490, opened 2025-09-02
Use `Vec<RegOrImmediate>` as const/global variable initializer

zluda-violet:
Also uses `RegOrImmediate` instead of a byte array for the initializer. Does not yet handle parsing registers in initializers.

### DIFF
--- a/ptx/src/pass/insert_implicit_conversions2.rs
+++ b/ptx/src/pass/insert_implicit_conversions2.rs
@@ -208,7 +208,7 @@ fn default_implicit_conversion_type(
         if should_bitcast(instruction_type, operand_type) {
             Ok(Some(ConversionKind::Default))
         } else {
-            Err(TranslateError::MismatchedType)
+            Err(error_mismatched_type())
         }
     } else {
         Ok(Some(ConversionKind::PtrToPtr))
@@ -264,14 +264,14 @@ pub(crate) fn should_convert_relaxed_dst_wrapper(
     (instruction_space, instruction_type): (ast::StateSpace, &ast::Type),
 ) -> Result<Option<ConversionKind>, TranslateError> {
     if operand_space != instruction_space {
-        return Err(TranslateError::MismatchedType);
+        return Err(error_mismatched_type());
     }
     if operand_type == instruction_type {
         return Ok(None);
     }
     match should_convert_relaxed_dst(operand_type, instruction_type) {
         conv @ Some(_) => Ok(conv),
-        None => Err(TranslateError::MismatchedType),
+        None => Err(error_mismatched_type()),
     }
 }
 
--- a/ptx/src/pass/llvm/emit.rs
+++ b/ptx/src/pass/llvm/emit.rs
@@ -247,10 +247,44 @@ impl<'a, 'input> ModuleEmitContext<'a, 'input> {
             unsafe { LLVMSetAlignment(global, align) };
         }
         if !var.array_init.is_empty() {
-            let initializer = get_array_init(self.context, &var.v_type, &*var.array_init)?;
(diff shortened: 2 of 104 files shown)


zluda-violet:
I'd missed the test failures; I'm looking into them.

zluda-violet:
The tests should now be passing so this is ready for review.

zluda-violet:
I've changed this to use variables in the `.reg` address space.

ptx/src/pass/llvm/emit.rs
```diff
@@ -445,7 +424,9 @@ impl<'a> MethodEmitContext<'a> {
             unsafe { LLVMSetAlignment(alloca, align) };
         }
         if !var.array_init.is_empty() {
-            todo!()
+            let initializer = get_array_init(self.context, &var.v_type, &var.array_init)?;
+            unsafe { LLVMBuildStore(self.variables_builder.get(), initializer, alloca) };
```
vosen: I thought that `LLVMSetInitializer` works on non-global variables, but LLVM says no (SPIR-V supports it https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpVariable).
The problem with the current approach is that we follow LLVM guidelines about `alloca`s and keep them in the separate entry basic block. This code breaks this by mixing `store`s into them. I leaving the right solution up to you. Two approaches come to mind:
* Remove initializer from non-global variables in earlier compiler passes and instead reintroduce constants (with initializers) (more effort, arguably nicer bitcode, not sure if it's worth it)
* In the line here use `builder` instead of `variables_builder` (less effort)

zluda-violet: I think I might switch back to using `ConstantDefinition`, since this approach is modifying `Variable` to add new behavior anyway and `ConstantDefinition` is more specific.

zluda-violet:
Okay, this should be ready for review again. Now it uses `ConstantDefinition` while adding support for registers in array initializers.

ptx/src/pass/insert_implicit_conversions2.rs
```diff
@@ -208,7 +208,7 @@ fn default_implicit_conversion_type(
         if should_bitcast(instruction_type, operand_type) {
             Ok(Some(ConversionKind::Default))
         } else {
-            Err(TranslateError::MismatchedType)
```
vosen, line 211: At certain point another pass uses `default_implicit_conversion_type` (or one of its callers), hence why it does not return a panicking error. Seems like this code has already been removed.
Since you are doing this cleanup please also remove all the `pub(crate)` from this file (well, except for the `run`)

ptx_parser/src/lib.rs
```diff
@@ -939,24 +951,26 @@ fn array_initializer<'b, 'a: 'b, 'input: 'a>(
...
                 ),
                 Token::RBrace,
             )
             .parse_next(stream)?;
             // pad with zeros
-            let result_size = type_.size_of() as usize * array_dimensions[0] as usize;
-            result.extend(iter::repeat(0u8).take(result_size - result.len()));
+            let result_size = array_dimensions[0] as usize;
+            result.extend(
+                iter::repeat(ast::RegOrImmediate::Imm(ast::ImmediateValue::U64(0)))
```
vosen: What about floating point immediates?
If they are handled correctly already then include them in tests

ptx/src/pass/insert_implicit_conversions2.rs
zluda-violet: Done.

ptx_parser/src/lib.rs
zluda-violet: Added test and handling for this.

PR #504, opened 2025-09-10
Random fixes

vosen:


vosen:
@zluda-violet this is a collection of random changes coming from the workload I'm working on. I want you to mainly take a look at the comiler changes made to make `param_is_addressable` pass. PTX uses `.param` namespace both for some local variables and kernel args. This is a problem for us because those are different address spaces on AMDGPU. So far we've made an effort to convert to local and const namespaces whenever possible, but this PR tries to handle some patterns which are impossible to track precisely by converting to generic space.

Side note: I don't quite remember why we don't coerce `.param` to generic address space, but there was a good reason for it. But I don't remember what it was specifically, so it might be worthwhile to revisit this approach in the future.

PR #512, opened 2025-09-16
Add support for cuBLASLt functions used by llm.c

zluda-violet:
Depends on #510


zluda_blaslt/src/impl.rs
```diff
@@ -18,6 +29,84 @@ impl ZludaObject for Handle {
 
 from_cuda_object!(Handle);
 
+pub struct MatmulDesc {
```
vosen: This whole wrapping of hipblasLtMatmulDesc_t, hipblasLtMatmulPreference_t and hipblasLtMatrixLayout_t seems unnecessary to me, if we wrapped those types with some extra data, but as it is now it's better to map them 1:1

vosen (changes requested):

zluda_blaslt/src/impl.rs
zluda-violet: Makes sense, done.

vosen (approved):

PR #527, opened 2025-09-25
No-op implementation of vprintf

zluda-violet:
This should be replaced later, possibly with an SSA pass.


ptx/src/pass/llvm/emit.rs
```diff
@@ -2853,6 +2854,7 @@ impl<'a> MethodEmitContext<'a> {
 
     fn emit_trap(&mut self) -> Result<(), TranslateError> {
         self.emit_intrinsic(c"llvm.trap", None, None, vec![])?;
+        unsafe { LLVMBuildUnreachable(self.builder) };
```
zluda-violet, line 2872: This is a separate bugfix that I forgot to separate out.

PR #528, opened 2025-09-25
Implement cuGraphExecUpdate_v2

zluda-violet:


zluda/src/impl/graph.rs
```diff
@@ -8,6 +9,48 @@ pub(crate) unsafe fn exec_destroy(graph_exec: hipGraphExec_t) -> hipError_t {
...
+pub(crate) fn exec_update_v2(
+    h_graph_exec: hipGraphExec_t,
+    h_graph: hipGraph_t,
+    result_info: &mut cuda_types::cuda::CUgraphExecUpdateResultInfo,
+) -> hipError_t {
+    let mut h_error_node: hipGraphNode_t = unsafe { std::mem::zeroed() };
+    let mut update_result: hipGraphExecUpdateResult = unsafe { std::mem::zeroed() };
+    unsafe { hipGraphExecUpdate(h_graph_exec, h_graph, &mut h_error_node, &mut update_result) }?;
+
+    result_info.errorNode = unsafe { std::mem::transmute(h_error_node) };
```
vosen: Is there a way to use `FromCuda` instead of transmute here and below?
In case we change the definition of CUgraphExec to be more complicated than transmute of `hipGraphExec_t`

zluda-violet: How's the new code look? It still uses `transmute` but in a wrapper struct returned from `FromCuda`

vosen:
This is very involved, how about something like this (with an appropriate comment why we do these acrobatics instead of `mem::transmute`):
```
    let error_node: *mut hipGraphNode_t = FromCuda::from_cuda(std::ptr::from_mut(&mut result_info.errorNode))?;
    let error_from_node: *mut hipGraphNode_t = FromCuda::from_cuda(std::ptr::from_mut(&mut result_info.errorFromNode))?;
    *error_node = h_error_node;
    *error_from_node = h_error_node;
```

zluda-violet:
Sure, done.

zluda/src/impl/graph.rs
```diff
@@ -8,6 +10,50 @@ pub(crate) unsafe fn exec_destroy(graph_exec: hipGraphExec_t) -> hipError_t {
...
 }
 
+pub(crate) fn exec_update_v2(
+    h_graph_exec: hipGraphExec_t,
+    h_graph: hipGraph_t,
+    result_info: &mut cuda_types::cuda::CUgraphExecUpdateResultInfo,
+) -> CUresult {
+    // We use FromCuda here instead of transmute in case our hipGraphNode_t representation changes
+    // in the future.
+    let mut h_error_node: hipGraphNode_t =
```
vosen: This looks wrong to me, my reading is that
* We load bits from `result_info.errorNode` into `h_error_node` on stack
* `hipGraphExecUpdate` writes into `h_error_node` on stack
* We copy bits from `result_info.errorNode` (without writing into it) into `result_info.errorFromNode`

zluda-violet: Yeah, you're correct. The code you wrote works I think, I just misread it before.

PR #530, opened 2025-09-30
Remove trailing whitespaces and add missing newlines at EOF

trueNAHO:
Although I am currently not using ZLUDA, I have been closely following its development for about the last 100 PRs to learn a few things. While reading the diffs for all this time, the missing newlines have been bothering me quite a bit, so I finally decided to open a PR for this. While at it, I also removed trailing whitespaces.

The `git diff` looks reasonable, although I have not tested anything.


docs/src/building.md
```diff
@@ -12,8 +12,8 @@
 
 ## Build steps
 
-* Git clone the repo (make sure to use `--recursive` option to fetch submodules):
```
vosen, line 15: Double spaces at the end here are intentional, that's how manual line breaks are done in markdown

vosen (changes requested):
Please reverse all changes in ext/detours. This code is third party and we want to keep it as-is to make updates easier.

I've noticed changes in one markdown file changing the output. Please review markdown files and their output.

Is there a way we can add those checks into our CI? Preferably with existing github action. We do have checks for formatting of Rust code, we can add more checks

docs/src/building.md
KubeRoot: Probably out of scope (and really unimportant), but the double spaces at the end should be unnecessary if you align the code line with the bullet point, like this:

```
* Git clone the repo (make sure to use `--recursive` option to fetch submodules):
  `git clone --recursive https://github.com/vosen/ZLUDA.git`
* Enter freshly cloned `ZLUDA` directory and build with cargo (this takes a while):
```

* Git clone the repo (make sure to use `--recursive` option to fetch submodules):
  `git clone --recursive https://github.com/vosen/ZLUDA.git`
* Enter freshly cloned `ZLUDA` directory and build with cargo (this takes a while):

trueNAHO: > Double spaces at the end here are intentional, that's how manual line breaks are done in markdown

This should be resolved in commit https://github.com/vosen/ZLUDA/pull/530/commits/8a6d7450124aef372c19a873fbe9af7d91b7319f.

.github/workflows/pr_master.yml
```diff
@@ -11,6 +11,30 @@ env:
...
+                --unified \
+                --label "a/$0" \
+                --label "b/$0" \
+                <(cat "$0") \
+                <(sed --regexp-extended 's/\s+$//; $a\') || status="$?"
+
+              if (($status != 1)); then
+                exit "$(($status + 1))"
+              fi
+            '
```
trueNAHO, line 34: > Is there a way we can add those checks into our CI? Preferably with existing github action. We do have checks for formatting of Rust code, we can add more checks

Commit https://github.com/vosen/ZLUDA/pull/530/commits/d5cf8efbd626029204764e5c39a8fa82c8dc6a9c should be able to do this. What is your preferred way of installing `fd` into the runner? Or is it already available?

vosen:
Thanks, I've pushed a minor change to install fd from Ubuntu repos and merged it

PR #551, opened 2025-10-29
Implement fallback MMA for RDNA1 and RDNA2

vosen:


vosen:
That one bf16 test fails purely because LLVM on ROCm 6.3.4 miscompiles it. Works fine on ROCm 6.4.4

PR #555, opened 2025-11-12
Build and distribute LLVM

zluda-violet:
We would like to add our own passes and instruction selection to LLVM, so this replaces comgr with building and calling LLVM ourself.

The most interesting new code is in zluda_llvm/compile.rs. It runs the `default<O3>` optimization pipeline and then generates code for the target architecture.


compiler/src/main.rs
```diff
@@ -135,11 +134,11 @@ fn report_pass_time(pass: &str, start: &mut Instant) {
     *start = Instant::now();
 }
 
-#[derive(Debug)]
 struct LLVMArtifacts {
-    bitcode: Vec<u8>,
+    context: llvm_zluda::utils::Context,
+    main: llvm_zluda::utils::Module,
```
vosen, line 138: Can you reorder `context` and `main`? There's probably a refcount or something that makes this drop order work fine, but I don't trust LLVM memory management and I'd like to drop Module before Context

llvm_zluda/src/compile.rs
```diff
@@ -0,0 +1,249 @@
...
+fn create_oclc_constants(ctx: &Context, gcn_arch: &str) -> Result<Module, String> {
+    let module = Module::new(ctx, c"oclc_constants");
+
+    // used by ockl
+    add_constant(ctx, &module, c"__oclc_wavefrontsize64", 0);
+    // This is set by Clang
+    // https://github.com/ROCm/llvm-project/blob/fb71c301a9a021ee5025b3ca7497970b4eb677df/clang/include/clang/Driver/RocmInstallationDetector.h#L21 and
+    // https://github.com/ROCm/llvm-project/blob/fb71c301a9a021ee5025b3ca7497970b4eb677df/clang/lib/Driver/ToolChains/CommonArgs.cpp#L3089
+    add_constant(ctx, &module, c"__oclc_ABI_version", 600);
+    add_constant(
```
vosen, line 76: This is slightly incorrect, ISA version reported by AMD runtime and used by ockl use different format.
E.g. runtime gfx1101 is __oclc_ISA_version 11001 (see in `amd\device-libs\oclc\src\isa_version_1101.cl`)

llvm_zluda/src/compile.rs
```diff
@@ -0,0 +1,249 @@
...
+
+fn create_oclc_constants(ctx: &Context, gcn_arch: &str) -> Result<Module, String> {
+    let module = Module::new(ctx, c"oclc_constants");
+
+    // used by ockl
+    add_constant(ctx, &module, c"__oclc_wavefrontsize64", 0);
+    // This is set by Clang
+    // https://github.com/ROCm/llvm-project/blob/fb71c301a9a021ee5025b3ca7497970b4eb677df/clang/include/clang/Driver/RocmInstallationDetector.h#L21 and
+    // https://github.com/ROCm/llvm-project/blob/fb71c301a9a021ee5025b3ca7497970b4eb677df/clang/lib/Driver/ToolChains/CommonArgs.cpp#L3089
+    add_constant(ctx, &module, c"__oclc_ABI_version", 600);
```
vosen: This should be aligned with LLVM, right now it work because LLVM builds for ABI v6 by default. We should explicitly pass `--amdhsa-code-object-version=6`

.gitmodules
```diff
@@ -1,7 +1,7 @@
 [submodule "ext/llvm-project"]
 	path = ext/llvm-project
-	url = https://github.com/llvm/llvm-project
-	branch = release/17.x
+	url = https://github.com/ROCm/llvm-project
+	branch = rocm-6.4.0
```
vosen: This is incorrect, `rocm-6.4.0` is a tag not a branch. In general `branch` field is only used when you automatically (with `git submodule update --remote`) update the submodule, it's fine to just remove it

vosen (changes requested):

llvm_zluda/src/compile.rs
vosen: ALso, change ABI to 5, version 6 does not bring anything interesting for us and only works with ROCm 6.3, which not everyone has

llvm_zluda/src/lib.cpp
```diff
@@ -216,4 +208,47 @@ void LLVMZludaSetAtomic(
...
+    std::string log_err_str;
+    llvm::raw_string_ostream log_out(log_out_str);
+    llvm::raw_string_ostream log_err(log_err_str);
+
+    // We don't want to leak memory as ZLUDA is running, so we destroy the linker context after every call, and wrap the invocation in a mutex.
+    std::lock_guard<std::mutex> guard(lld_mutex);
+    lld::Result result = lld::lldMain(args, log_out, log_err,
+                                      {{lld::Gnu, &lld::elf::link}});
+
+    lld::CommonLinkerContext::destroy();
```
vosen: Do we really need this line?
It looks bit disconnected from the rest of the code and looking at LLD sources: https://github.com/llvm/llvm-project/blob/630dfc928c1e0668c3c0d59b303bb968ba44d7dd/lld/Common/DriverDispatcher.cpp, `lldMain` calls the function by itself anyway

llvm_zluda/src/lib.cpp
```diff
@@ -216,4 +208,47 @@ void LLVMZludaSetAtomic(
...
+
+int LLVMZludaLinkWithLLD(const char *input_path, const char *output_path, char **ErrorMessage)
+{
+    std::vector<const char *> args;
+    args.push_back("ld.lld");
+    args.push_back("-shared");
+    args.push_back(input_path);
+    args.push_back("-o");
+    args.push_back(output_path);
+    args.push_back("--threads=1");
```
vosen, line 221: Please also add `--no-undefined`. That flag is important to make sure that we fail compilation if we are linking to non-existing function instead of explodingin various undebuggable way

vosen (changes requested):
I've read the code. Good changes, but some fixes necessary

compiler/src/main.rs
zluda-violet: Done (I moved attributes as well).

llvm_zluda/src/compile.rs
zluda-violet: Oh, right. Fixed.

llvm_zluda/src/compile.rs
zluda-violet: Done.

.gitmodules
zluda-violet: Done.

llvm_zluda/src/lib.cpp
zluda-violet: You're correct, we don't.

llvm_zluda/src/lib.cpp
zluda-violet: Done.

vosen (commented):
Looks good to me, but some tests are failing

PR #560, opened 2025-11-17
Small fix for Rust-CUDA support

stevefan1999-personal:
Bare minimum needed for https://github.com/Rust-GPU/rust-cuda to run the vecadd example.

Pushes #561 a little further


zluda_common/src/lib.rs
```diff
@@ -207,6 +208,7 @@ from_cuda_nop!(
 from_cuda_transmute!(
     CUuuid => hipUUID,
     CUfunction => hipFunction_t,
+    CUjit_option => hipJitOption,
```
vosen: As far as I can tell this is not used in the code, `load_data_ex` uses `CUjit_option` and not `hipJitOption`. 
Transmute is fine if the types have the same bit representation, but it looks like at a certain point HIP stopped caring:
```
pub const hipJitOptionPositionIndependentCode: hipJitOption = hipJitOption(25);
```
```
pub const CU_JIT_POSITION_INDEPENDENT_CODE: CUjit_option_enum = CUjit_option_enum(
        30,
    );
```

I think it's better to remove this line

vosen:
Merged, thank you

PR #574, opened 2025-12-08
Additional NVML functionality

zluda-violet:
Implements `nvmlDeviceGetMemoryInfo` and `nvmlDeviceGetName` using ROCm SMI on Unix, and `nvmlDeviceGetCudaComputeCapability` using a constant version for all devices (the CUDA version that ZLUDA emulates).

### DIFF
--- a/zluda_ml/src/impl_unix.rs
+++ b/zluda_ml/src/impl_unix.rs
@@ -180,7 +180,7 @@ pub(crate) unsafe fn device_get_memory_info(
 
     *memory = nvmlMemory_t {
         total,
-        free: total - used,
+        free: total.saturating_sub(used),
         used,
     };
 


zluda-violet:
This is part of the vLLM work. Does vLLM use nvml on Windows? If so we're going to need to figure out how to handle that.

vosen:
Yes, nvml is used on Windows too. The problem is as usual AMD.
AMD for a long time refused to port rocm-smi to Windows. Now, they have decided to replace rocm-smi with a fresh new amdsmi. That is also Linux-only.
On Windows there are ADL and ADLX which do some of the things that rocm-smi does. But while rocm-smi is fairly straightforward, ADL/ADLX are janky and cryptic. I were holding out in hope of a better library on Windows, but that does not seem likely anytime soon.
For now let's focus on Linux, but yes, we will have to do the Windows side soon

PR #580, opened 2025-12-15
Allow implicit conversion from vec to bit scalar for ld

zluda-violet:
Fixes error from code like:

```
ld.b64   { output1, output2 }, [input];
```

### DIFF
--- a/ptx/src/pass/test/mod.rs
+++ b/ptx/src/pass/test/mod.rs
@@ -174,6 +174,10 @@ impl<'a> ast::VisitorMap<SpirvWord, SpirvWord, ()> for StatementFormatter<'a> {
             {
                 let type_string = if let Some((type_, state_space)) = type_space {
                     // We use the type_space from the resolver rather than from the operand, to avoid hiding implicit conversions
+                    let state_space = match state_space {
+                        ast::StateSpace::Generic => ".generic".to_string(),
+                        _ => format!("{}", state_space),
+                    };
                     format!("{}{} ", type_, state_space)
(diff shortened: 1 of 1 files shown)


ptx/src/pass/test/insert_implicit_conversions/default_relaxed.ptx
```diff
@@ -0,0 +1,24 @@
...
+    ret;
+}
+
+// %%% output %%%
+
+.func (.reg .b32 %2, .reg .b32 %3) %1 (
+    .reg .b64 %4
+)
+{
+    .b64 %7 = zluda.convert_implicit.bit_to_ptr.b64.reg.b64 %4;
```
vosen, line 19: I guess after the conversion ZLUDA tracks `%7` as a variable of type `.b64` in `.generic` space? In this case, would it be possible to have this line instead be as follows do readability? Just for variables that are in non-`.reg` address space.
`.generic.b64 %7 = zluda.convert_implicit.bit_to_ptr.b64.reg.b64 %4;`

vosen (commented):
Lookgs good, one nit about readability

ptx/src/pass/test/insert_implicit_conversions/default_relaxed.ptx
zluda-violet: Sure. I've added `.generic`, I'll swap the type and state space to be consistent with PTX variable declaration (`.state_space.type`) in a follow-up.

PR #585, opened 2025-12-17
Allow implicit conversion from bit scalar to vec for st

zluda-violet:
The inverse of #580, for `st` instead of `ld`.


vosen:
Implicit conversion code is quite messy, long term we should throw it all away and have a much more structured and less ad-hoc approach starting with tests for every possible pair of types and all possible instruction classes: `mov`, `st`, `ld`, normal instructions

PR #595, opened 2025-12-31
Implement cuMemHostGetDevicePointer_v2

Knogle:
**Note: Wasn't able to check Violet's approach yet!**

This change adds a full implementation of cuMemHostGetDevicePointer_v2 by mapping it to hipHostGetDevicePointer, enabling proper support for CUDA host-mapped memory allocations (CU_MEMHOSTALLOC_DEVICEMAP) under ZLUDA on Windows. Cinebench and other CUDA applications that rely on pinned + device-mapped host memory now progress instead of returning CUDA_ERROR_NOT_SUPPORTED.

Notes:
- Enables MemCpyKernelUInt lookup via symbol fallback
- Introduces no functional regressions in existing flows

### DIFF
--- a/zluda/src/impl/memory.rs
+++ b/zluda/src/impl/memory.rs
@@ -119,7 +119,7 @@ pub(crate) unsafe fn host_get_device_pointer_v2(
     }
 
     // HIP equivalent of cuMemHostGetDevicePointer_v2
-    hipHostGetDevicePointer((&mut pdptr.0 as *mut *mut ::core::ffi::c_void).cast(), p, flags)?;
+    hipHostGetDevicePointer(std::ptr::from_mut(pdptr).cast(), p, flags)?;
     Ok(())
 }
 


zluda/src/impl/memory.rs
```diff
@@ -109,6 +109,20 @@ pub(crate) unsafe fn host_alloc(
...
+    pdptr: &mut hipDeviceptr_t,
+    p: *mut ::core::ffi::c_void,
+    flags: ::std::os::raw::c_uint,
+) -> CUresult {
+    if p.is_null() {
+        return CUresult::ERROR_INVALID_VALUE;
+    }
+
+    // HIP equivalent of cuMemHostGetDevicePointer_v2
+    hipHostGetDevicePointer((&mut pdptr.0 as *mut *mut ::core::ffi::c_void).cast(), p, flags)?;
```
vosen: This is correct, but for better clarity I would prefer this:
```
hipHostGetDevicePointer(std::ptr::from_mut(pdptr).cast(), p, flags)?;
```

vosen (changes requested):

zluda/src/impl/memory.rs
Knogle: Done so :)

PR #607, opened 2026-01-16
Implement extended precision integer addition

zluda-violet:
Uses `llvm.*add.with.overflow.*`. That intrinsic does not take a carry argument, so handling carry in requires multiple additions and combining the carry out, but the AMDGPU target is able to translate that pattern into a single instruction.

These four PTX instructions:

```
    add.cc.u32      r0, a0, b0;
    addc.cc.u32     r1, a1, b1;
    addc.cc.u32     r2, a2, b2;
    addc.u32        r3, a3, b3;
```

are translated into four RDNA3 instructions:

```
    v_add_co_u32 v0, vcc_lo, v0, v4
    v_add_co_ci_u32_e32 v1, vcc_lo, v1, v5, vcc_lo
    v_add_co_ci_u32_e32 v2, vcc_lo, v2, v6, vcc_lo
    v_add_co_ci_u32_e32 v3, vcc_lo, v7, v3, vcc_lo
```

### DIFF
--- a/ptx/src/pass/llvm/emit.rs
+++ b/ptx/src/pass/llvm/emit.rs
@@ -915,7 +915,10 @@ impl<'a> MethodEmitContext<'a> {
         Ok(carry_flag)
     }
 
-    fn emit_load_carry_flag(&mut self, type_: ast::ScalarType) -> Result<LLVMValueRef, TranslateError> {
+    fn emit_load_carry_flag(
+        &mut self,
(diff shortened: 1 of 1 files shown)


zluda-violet:
A bit simpler yeah, I'm mostly just avoiding updating all the tests to include the carry flag.

vosen:
Makes sense

PR #609, opened 2026-01-20
Finish extended precision arithmetic

zluda-violet:
Adds `sub.cc`, `subc`, `subc.cc`, `mad.cc`, `madc`, and `madc.cc` on top of the work done in #607.


zluda-violet:
LLVM might not be producing the optimal instructions for this code, I'm looking into why this seems to behave differently than addition.

zluda-violet:
I've fixed the issue with sub-optimal assembly by cherry-picking this PR: https://github.com/vosen/llvm-project/pull/3.

PR #647, opened 2026-06-08
ptx: parse `exit` instruction

Allen050329:
The parser had no rule for `exit`, so any kernel using it failed with `UnrecognizedStatement("exit;")`. Because directives are parsed with error recovery, that drops the *entire enclosing kernel*; in release builds the module loader uses `parse_module_unchecked`, which accepts a module with `invalid_directives > 0`, so the kernel is silently absent and a later `cuModuleGetFunction` / `cuLibraryGetKernel` returns NOT_FOUND.

This bites real code: CUB's `DeviceRadixSort` onesweep kernel early-outs an out-of-range tile with `exit`, so `DeviceRadixSort` silently misbehaves under ZLUDA (the sort never runs and callers that don't check the return code get unsorted data).

`exit` terminates the calling thread regardless of call depth. That matches a void return only in a top-level entry routine (PTX ISA: "A return instruction executed in a top-level entry routine will terminate thread execution"); inside a callable `.func` a `ret` would unwind to the caller and let the thread keep running. ZLUDA does not inline `.func`s, so parse `exit` to its own `Instruction::Exit` terminator (handled like `Trap` through the passes) and at emit lower it to `Ret` when the enclosing method is an entry routine, otherwise raise a translate error rather than silently mis-executing.

Adds a parser unit test and a spirv_run test that early-outs with `exit` in an entry kernel and checks the post-exit store does not execute.

### DIFF
--- a/ptx/src/pass/insert_post_saturation.rs
+++ b/ptx/src/pass/insert_post_saturation.rs
@@ -202,6 +202,7 @@ fn run_instruction<'input>(
         }
         | ast::Instruction::Tanh { .. }
         | ast::Instruction::Trap {}
+        | ast::Instruction::Exit {}
         | ast::Instruction::Xor { .. }
         | ast::Instruction::Vote { .. }
         | ast::Instruction::ReduxSync { .. }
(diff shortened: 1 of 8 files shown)


vosen:
I'm closing it in favor of #643. It already contains implementation of `exit` (and some other things). I just forgot to merge it.
Sorry for that

PR #651, opened 2026-06-26
32 bit CUDA support

vosen:
This is far from ready, but if one is inclined to try it the steps are. Just remember **you do this at your own risk**:
* Get source code
* Adjust hard-coded paths in `zluda32\src\ipc.rs` from `C:\dev\ZLUDA` to whatever is your path
* Build for 64 bit and 32bit: `cargo xtask -r` and `cargo xtask --target=i686-pc-windows-msvc -r`
* Copy 32 bit nvapi.dll and nvcuda.dll from `target\i686-pc-windows-msvc\release` to `C:\Windows\SysWOW64`
* Launch your app, it might work or not


kcajjones:
Does this allow a secondary GPU to be dedicated to Physx? Secondary CPU being AMD or Nvidia.

vosen:
> Does this allow a secondary GPU to be dedicated to Physx? Secondary CPU being AMD or Nvidia.

It depends on your setup. Right now it runs on the first AMD GPU it sees. If you have dual AMD GPU it might pick a GPU you don't want, I'll probably allow users to control which GPU to pick. If you have AMD + NV setup you can control by picking which PhysX implementation you use: NVIDIA or ZLUDA

niobium93:
Is it possible to cross-compile the 32bit libs and 64bit server on linux for windows? Could they run inside wine in conjunction with SveSop's [nvidia-libs](https://github.com/SveSop/nvidia-libs) / [nvcuda](https://github.com/SveSop/nvcuda/) for wine-to-native-cuda bindings? Would this need special support in [dxvk-nvapi](https://github.com/jp7677/dxvk-nvapi) ?

vosen:
You should be able to make it run in WINE. It depends on how WINE interacts with the Linux host.
They way it works right now on Windows is that there's three pieces:
* 32-bit ZLUDA dll
* 64-bit ZLUDA server executable
* 64-bit ZLUDA dll

PhysX workload loads 32-bit ZLUDA dll, which starts 64-bit ZLUDA server process and communicates with it through RPC (named pipe IIRC). 64-bit ZLUDA server loads is a proxy 64-bit ZLUDA dll and acts as a proxy, maintaing the fake 32-bit state and translating 32-bit ZLUDA driver calls to 64-bit. The problem is that all three are in cahoots. 64-bit ZLUDA dll knows that the GPU kernels it compiles are compiled for a special 32-bit on 64-bit mode. 64-bit server knows that 64-bit ZLUDA dll is in special mode and can query for some extra metadata to successfully launch kernels and maintain internal state.

When it come to OS dependencies it all relies on WoW64 working properly (launching 64-bit exe from 32-bit exe and being to RPC to it) and having access to ROCm runtime (amdhip64_6.dll/amdhip64_7.dll). As long as those two requirements are met everything should more or less work ok. There's obviously no native ROCm on WINE, but there should be a way to expose Linux ROCm runtime as WINE Windows ROCm runtime, I'd focus on that.

niobium93:
There isn't anything that proxies ROCm/HIP through WINE to native Linux ROCm as far as I know, but there is a proxy library for CUDA. Does the 64bit proxy library load the real nvcuda.dll, or does it contain ZLUDA's nvcuda implementation? For example is it possible to use this atop NVIDIA's own CUDA implementation? My plan would be to use ZLUDA 32-to-64 atop SveSop's nvcuda proxy and then have native ZLUDA on the Linux side.

vosen:
It won't work with NVIDIA's own CUDA

Nuck-TH:
Don't know if it is appropriate place for this, but anyway:
Built this, placed relese and i686-pc-windows-msvc from target to C:\dev\ZLUDA\target, copied dlls, and tried to run Alice: Madness Returns, but no dice - it falls back to cpu, occasionally crashes on level loading(physx is apparently intialised only then). rpc server starts on game start, so at least that seems to be working.
About crashes - running zluda.exe -- cuda_check.exe randomly crashes, though not always. Windows 10, HIP 6.4.0, rx 6750xt with latest driver.

If needed i can describe build environment.

Here is logs created by running game via i686 zluda.exe with --nvidia-trace and then --zluda-trace cmd parameters:
https://gist.github.com/Nuck-TH/aecafb6215060a9a697da123d5dafc81

Can you give links to that tech demos from blog post? - it would help to check if my build on my system is working at all.

vosen:
@Nuck-TH
I am not quite following what you did with the dlls, but it seems to be fine. Your workload is failing because it uses `cuCtxCreate`, which was not implemented (I've added it now). Probably an older version of physx.

You can download PhysX SDK samples from here: https://developer.download.nvidia.com/PhysX/2.8.1/PhysX_2.8.1_SDK_Core.msi

You can also try fluidmark (support added today): https://www.geeks3d.com/20130308/fluidmark-1-5-1-physx-benchmark-fluid-sph-simulation-opengl-download/

Nuck-TH:
Next bunch(Alice):
https://gist.github.com/Nuck-TH/3536fb55103d8303025db233a4c584d4
According to pcgamingwiki and file versions it uses physx 2.8.4.4

Do i need both logs or one(which) is sufficient?

Fluidmark gets stuck at initializing fluid 0%, nvidia samples segfault. If fluidmark is started any(x64/i686) zluda.exe - it errors out with "could not load particle texture".
Seems like i'm doing something wrong after all :( Figures what...

Also, for some reason, zluda stopped creating cahe file, even if i run precompile...

> I am not quite following what you did with the dlls

Did what is told in OP

> Copy 32 bit nvapi.dll and nvcuda.dll from target\i686-pc-windows-msvc\release to C:\Windows\SysWOW64

RunnerFaith:
Managed to get Mirror's Edge working with this - followed the steps provided here and then **all** the steps under the "Install and enable the latest PhysX update" section on [PCGamingWiki](https://www.pcgamingwiki.com/wiki/Mirror%27s_Edge#cite_ref-32) so that the game uses the newer version of PhysX and loads the ZLUDA DLLs along with it.

Settings used:
Mirror's Edge game settings - all settings maxed @ 1440p with 8XQ anti-aliasing
**FPS limit removed using Mirror's Edge Tweaks.**

Run-through of chapter 1 (checkpoint B): https://youtu.be/_wMEfFz_KLQ
CPU PhysX run of the same level for comparison: https://youtu.be/Lni8RE-mr60

Notes:
- Performance when glass breaks is far better with ZLUDA vs CPU PhysX. For example, the scene @ 3:21 is basically a slideshow with CPU PhysX at single digit frames compared to ZLUDA. **See the comparison video above for more.**
- ~~Stuttering @ 0:38 seems to be present only when using ZLUDA. The other stutters @ 1:48 & 3:45 seem to happen regardless, I assume this is a game issue but will double check with PhysX off.~~ Disregard, this is unrelated to ZLUDA.
- Least performant areas look to be the chopper scene @ 1:00 & the elevator hallway scene @ 2:20 - I assume this is down to multiple cloth simulations in these scenes (blinds/shades & tarp respectively) but not 100% sure

zluda_trace compressed log.txt (different run from the recording but same level): [zluda_trace.zip](https://github.com/user-attachments/files/29676277/zluda_trace.zip) - I can provide the full log output if needed.

Nuck-TH:
This is what i get after just installing [this](https://www.nvidia.com/object/physx_system_software.html) and moving to HIP nightlies
(unlike for mirrors edge i didn't replace filed in game folder as it refuses to start with that)
[AliceMadnessReturns.exe.zip](https://github.com/user-attachments/files/29676622/AliceMadnessReturns.exe.zip)

This got FluidMark and nvidia samples working :)

R4to0:
> Managed to get Mirror's Edge working with this - followed the steps provided here and then steps 2 & 3 of the "Install and enable the latest PhysX update" section on [PCGamingWiki](https://www.pcgamingwiki.com/wiki/Mirror%27s_Edge#cite_ref-32) so that the game loads the ZLUDA DLLs.

Thank you! Worked fine on my end following your instructions.

Here's some recordings:
ZLUDA OFF

https://github.com/user-attachments/assets/68aac199-23fd-488c-b453-76c8009bdf67

ZLUDA ON

https://github.com/user-attachments/assets/f0a54449-47ff-42ae-bca8-92d508258665


Hardware:
AMD Radeon RX 6600 8GB
AMD Ryzen 5 3600X
32GB DDR4@3200MT/s

Edit: test with Batman Arkham Origins, framerate wasn't that bad running on the CPU but was a lot choppy too.

ZLUDA OFF

https://github.com/user-attachments/assets/7778acba-766d-45f8-a16e-8e2fa6654586



ZLUDA ON

https://github.com/user-attachments/assets/6cc0e541-21ad-4f3b-9fdf-ba113fd946d7

vosen:
@Nuck-TH try the most recent commit, it should work with Alice: Madness Returns now

As for performance, in general there's not much improvement to be made. Maybe by redoing the communication between 32-bit ZLUDA and 64-bit server from named pipes to shared memory. But that's a bit tricky.

Thanks for testing it, I did not consider installing latest physx system-wide (I already had it installed by the NV drivers). I'll be focusing on cleaning it up for merge and updating documentation (turns out you don't need to put nvcuda.dll in syswow64, you can just launch from steam by setting the command to `"<PATH_TO\zluda.exe>" -- %command%`), but if you have more games feel free to upload logs here before the merge

Nuck-TH:
After this changes Alice actually works \\^_^/

Though some effects lower fps still, which is odd for a game that uses phys for very few select things. Seems like RPC is bottleneck after all, because compute load doesn't rise beyond certain amount(same as with FluidMark). Probably implementation of integration into engine and/or effects is bad and spams too much calls. I wouldn't be surprised, because it works horribly on CPU.

But there is peculiarity with zluda.exe  - if i run the game via 64bit one i don't get any trace log. If i run with 32bit one - physx fails exact same way as in [comment #4885364121](https://github.com/vosen/ZLUDA/pull/651#issuecomment-4885364121)... So i can't capture trace of things working.

Another thing is that there are issues with reinitializing physics - in fluidmark if you start and abort test/benchmark a few times it eventually fails and falls back to cpu. Same with nvidia samples(like SampleCloth.exe) that can switch scenes or toggle hardware acceleration.

vosen:
Running with 64 bit zluda.exe will do nothing, you need to run 32 bit zluda.exe: those two are different and each only works with their corresponding bitness. I want to eventually provide a single loader that supports any bitness, but that will have to wait.

The best way to run it is to change your Steam launch options to `"<PATH_TO_I686\zluda.exe>" -- %command%` or, if you want logs, to `"<PATH_TO_I686\zluda.exe>" --zluda-trace -- %command%`. Or if it's not on Steam, run directly with `<PATH_TO_I686\zluda.exe> -- <PATH_TO_GAME_EXE>`

As for reinitialization I agree it's a problem, I'll work on it after merging this PR.

Thanks for testing

Nuck-TH:
> Running with 64 bit zluda.exe will do nothing, you need to run 32 bit zluda.exe: those two are different and each only works with their corresponding bitness.

As i written - i figured as much.

> or, if you want logs, to "<PATH_TO_I686\zluda.exe>" --zluda-trace -- %command%

I know, yet:

> If i run with 32bit one - physx fails exact same way as in https://github.com/vosen/ZLUDA/pull/651#issuecomment-4885364121... So i can't capture trace of things working.

Physx works if game is run by itself, but not if via 32bit zluda executable.

vosen:
If it's a Steam game, it doesn't work if you start it through zluda.exe outside of Steam. The game executable instead of running normally goes into special mode where it messages Steam to do the actual launch and it closes. Then Steam relaunches it, totally separately and outside of zluda.exe control

pchome:
Looks promising. Linux, WINE basic wow64 unixlib.
zluda_gpu_caps_pre
[log.txt](https://github.com/user-attachments/files/29714478/log.txt)

vosen:
Now that it has been merged, please move follow-up discussion to Issues/Discussion in this repo or to Discord.
#656 contains the list of things I plan to do to improve/fix this feature

PR #654, opened 2026-07-01
docs: clarify unofficial HIP SDK column refers to AMD nightlies

0xDELUXA:
The tarballs in the right-hand column are produced by [TheRock](https://github.com/ROCm/TheRock), AMD's own open-source build system, so both "unofficial" and "unsupported by AMD" wrongly implied AMD isn't involved.

This reframes the contrast as stable release vs nightly  while keeping the honest warning that nightlies carry no stability guarantees.


vosen:
Thank you

PR #661, opened 2026-07-23
Improve floating point accuracy in complex kernels

vosen:
This PR will fix two issues:
* Minor problem where some reason I've been using wrong intrinsic for floating point minimum/maximum, which could theoretically lead to miscompute
* Major issue where ZLUDA didn't emit constrained intrinsics and that can lead, in certain complex cases, to emission of floating point instructions with wrong rounding modes. Most of the work is actually on the LLVM side, but there's a lot of busywork in the Rust compiler too


vosen:
llvm side is in vosen/llvm-project/pull/14

Issue #652, opened 2026-06-28
Implement multi-threaded compilation


vosen:
Expected outcome:
Compiling large PTX module uses multiple threads

Comments:
* ZLUDA passes are all relatively fast, so there's little reason to divide it between threads, all the time is spent in LLVM
* Blender is a good target for experiments. They ship single PTX module, it's 12MB of pure text and contains 75 kernels 
* The strategy would be to write a transformation that splits PTX module into separate compilation units, one for each kernel and than compiling them in parallel and linking into a single ELF
* Extra care has to be applied to non-kernel top-level items: shared memory globals, global memory globals, functions. AMDGPU target and lld have full support for linking, but it will require some care to make sure that linking/visibility attributes are emitted correctly and  things link properly

0rlych1kk4:
Hi @vosen  I am interested in investigating this issue. Before implementing the full transformation, I would like to trace the current PTX-to-AMDGPU compilation path and prototype splitting a simple multi-kernel PTX module into separate LLVM compilation units. I would then verify handling of shared globals and device functions, link the resulting objects with lld, and compare correctness and compilation time against the current single-module path. Would this be a suitable initial direction, and is there a preferred part of the compiler pipeline where you expect the split to occur?

vosen:
yes, that's the goal. There's three reasonable places the split could happen for various reasons:
* Just after parsing, because that's the first spot where one can do anything meaningful
* Just after converting all text symbols to `SpirvWord` (`normalize_identifiers`), because it's convenient to work with the code (but you need to clone the resolver, which might be costly?)
* Just before emitting llvm, which is the latest possible point. All the compilation time is in the llvm

You can get pass timing by compiling with zoc (`compiler` crate)

0rlych1kk4:
Thanks @vosen  that gives me a clear starting point. I’ll first build with zoc and check the pass timings, then trace the state just before LLVM emission. I’ll prototype the split there first, since it keeps parsing and identifier normalization unchanged and focuses directly on the expensive part of the pipeline. I’ll start with a small multi-kernel PTX case before testing anything larger like Blender.

PR #662, opened 2026-07-23
Split PTX LLVM emission into common and per-kernel modules

0rlych1kk4:
## What this changes

This PR splits PTX LLVM emission into a common module and separate modules for each kernel.

The common module keeps the global variables and helper function definitions. Each kernel module contains the kernel itself, declarations for the helper functions it calls, and extern declarations for the globals it may reference. The kernel modules are then linked back into the common module.

## Why

The current emission path builds one LLVM module from all transformed PTX directives. This change introduces the structure needed to compile kernels separately without cloning the transformed instruction bodies.

## Implementation notes

- tracks direct and transitive helper dependencies for each kernel;
- creates body-less declarations for reachable helper functions;
- builds an ownership-based compilation plan so definitions are moved only once;
- creates extern global declarations for kernel modules;
- emits each kernel module separately and links it into the common module.

The monolithic reconstruction path is kept only for tests.

## Validation

I ran:

- `cargo check -p ptx`
- `cargo check -p ptx --tests`
- `cargo build -p compiler --bin zoc`

I also compiled these PTX files through LLVM linking and ELF generation:

- `ptx/src/test/spirv_run/add.ptx`
- `ptx/src/test/spirv_run/call.ptx`
- `ptx/src/test/spirv_run/global_array.ptx`
- `ptx/src/test/spirv_run/extern_shared_call.ptx`
- `ptx/src/test/spirv_run/32/add.ptx`

The generated LLVM IR retained the helper definitions, global initializers, external shared declarations, and kernel calls as expected.
Opening this as a draft to get feedback on the module ownership and linking approach before expanding the test coverage.

## Part of #652 
This implements the compilation-unit splitting and linking groundwork for #652.

Each kernel is now emitted as a separate LLVM module with the required helper and global declarations, then linked into the common module. The current implementation still emits the kernel modules sequentially. Parallel scheduling of those compilation units is not included in this PR.

### DIFF
--- a/ptx/src/pass/kernel_dependencies.rs
+++ b/ptx/src/pass/kernel_dependencies.rs
@@ -118,3 +118,58 @@ pub(super) fn kernel_declaration_sets(
         })
         .collect()
 }
+
+pub(super) struct KernelModulePlan {
+    pub(super) kernel: Function<ast::Instruction<SpirvWord>, SpirvWord>,
+    pub(super) declarations: Vec<Function<ast::Instruction<SpirvWord>, SpirvWord>>,
+}
+
+pub(super) struct KernelCompilationPlan {
+    pub(super) common: Vec<Directive2<ast::Instruction<SpirvWord>, SpirvWord>>,
+    pub(super) kernels: Vec<KernelModulePlan>,
+}
+
+pub(super) fn build_compilation_plan(
+    directives: Vec<Directive2<ast::Instruction<SpirvWord>, SpirvWord>>,
+) -> KernelCompilationPlan {
+    let mut declaration_sets = kernel_declaration_sets(&directives);
+    let mut common = Vec::new();
+    let mut kernels = Vec::new();
+
+    for directive in directives {
+        match directive {
+            Directive2::Variable(..) => common.push(directive),
+            Directive2::Method(function) if function.is_kernel() => {
+                let declarations = declaration_sets.remove(&function.name).unwrap_or_default();
+
+                kernels.push(KernelModulePlan {
+                    kernel: function,
+                    declarations,
+                });
+            }
+            Directive2::Method(..) => common.push(directive),
+        }
+    }
+
+    KernelCompilationPlan { common, kernels }
+}
+
+impl KernelCompilationPlan {
+    pub(super) fn into_monolithic_directives(
+        self,
+    ) -> Vec<Directive2<ast::Instruction<SpirvWord>, SpirvWord>> {
+        let mut directives = self.common;
+
+        directives.extend(self.kernels.into_iter().map(|kernel_plan| {
+            let KernelModulePlan {
+                kernel,
+                declarations,
+            } = kernel_plan;
+
+            drop(declarations);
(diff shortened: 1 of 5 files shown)


0rlych1kk4:
Hi @vosen   separately from the PR, I’m currently looking for paid work involving Rust, LLVM, GPU tooling, or compiler development. I’ve enjoyed working on this part of ZLUDA and would be interested in contributing professionally if there is a suitable opportunity, including contract, part-time, or full-time work. I’d also appreciate any referral or direction you may have. Thank you.

vosen (changes requested):
Code in `kernel_dependencies.rs` has to be heavily reworked

* it ignores the dependencies in global variables:
  ```
  const  .u32 foo = 42;
  .global .u32 p2 = foo;
  ```

* It seems to ignore existence of declarations (going by `FxHashMap<SpirvWord, &Function<ast::Instruction<SpirvWord>, SpirvWord>>`):
  ```
  .extern .func foobar(.b64 x );
  
  ...
  
  .func foobar(.param .b64 x) {
      ...
  }
  ```
  Every function can be identified by SpirvWord, but there might be multiple Function objects corresponding to a single function (because of declarations)

* The style is weird, why does each iterator one-liner need its own function? And why the function ordering is C++ style?

The code should first build a reachability graph (petgraph is already a dependency) and then split it into separate modules and compile

I did not read tests in detail, but they can be simplified too. Either by introducing some helpers or by extending existing text assertions code (see tests in `ptx/src/pass/test`)

vosen:
> Hi @vosen separately from the PR, I’m currently looking for paid work involving Rust, LLVM, GPU tooling, or compiler development. I’ve enjoyed working on this part of ZLUDA and would be interested in contributing professionally if there is a suitable opportunity, including contract, part-time, or full-time work. I’d also appreciate any referral or direction you may have. Thank you.

Sorry, but I know of no such opportunities currently

0rlych1kk4:
Thanks for the detailed feedback. I understand the concerns around global-variable dependencies, multiple declarations for the same function, the current structure of kernel_dependencies.rs, and the test style.
I’ll rework this around a proper reachability graph using petgraph, account for declarations and global dependencies, and simplify the tests using the existing test helpers where possible. I’ll update the PR once the revised approach is ready.
And thank you as well for answering my question about work opportunities.

0rlych1kk4:
@vosen , I’ve reworked this based on your comments.
The dependency analysis now builds a petgraph reachability graph covering function calls, global references from function bodies, and dependencies between global initializers. Function declarations and definitions with the same SpirvWord are retained instead of one replacing the other.
The compilation plan now uses that reachability result to emit only the required function and global declarations for each kernel, while preserving directive order.
I also updated the tests for declaration/definition handling, global initializer dependencies, function-to-global references, unreachable globals, and declaration ordering.
cargo check -p ptx and cargo check -p ptx --tests both pass locally.

0rlych1kk4:
@vosen  , I’ve also rebased the branch onto the current master and resolved the conflicts. The updated branch is ready for re-review.

PR #665, opened 2026-08-02
Enable Windows Job Objects APIs

AmosKito1:
## Summary

Declare the Windows API feature groups required by ZLUDA's existing Job Objects
calls, and test the Windows support crate independently in pull-request CI.

## Root cause

`zluda_windows` uses Job Object APIs but did not declare the corresponding
Windows crate feature groups. Full workspace builds can hide this because Cargo
unifies features requested by other crates; an independent `zluda_windows`
build then fails to compile.

## Impact

Restores standalone Windows builds of `zluda_windows` and prevents this
dependency-feature regression from being masked by unrelated workspace crates.

## Validation

- `cargo test -p zluda_windows --lib --locked`
  - 6 passed

### DIFF
--- a/.github/workflows/pr_master.yml
+++ b/.github/workflows/pr_master.yml
@@ -79,8 +79,6 @@ jobs:
         submodules: true
     - name: Run sccache-cache
       uses: mozilla-actions/sccache-action@v0.0.9
-    - name: Test Windows support crate
-      run: cargo test -p zluda_windows --lib
     - name: Build
       run: |
         cargo xtask zip --profile ${{ env.CARGO_PROFILE }}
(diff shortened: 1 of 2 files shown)


.github/workflows/pr_master.yml
```diff
@@ -79,6 +79,8 @@ jobs:
         submodules: true
     - name: Run sccache-cache
       uses: mozilla-actions/sccache-action@v0.0.9
+    - name: Test Windows support crate
```
vosen: We don't have Windows runners set up and anyway the tests in this crate are unlikely to fail, that crate doesn't get any activity that would invalidate them. Please remove this

vosen (changes requested):
Looks good, one change needed

vosen (approved):

PR #668, opened 2026-08-10
Mask dynamic shuffle delta to five bits

AmosKito1:
Implements PTX-specified five-bit masking for dynamic `shfl.sync` delta operands.

This matches CUDA's `b[4:0]` shuffle operand semantics for all four shuffle modes.

Tests cover dynamic values 32, 33, and 63 for `up`, `down`, `bfly`, and `idx`, including returned values and predicates.


ptx/src/test/spirv_run/mod.rs
```diff
@@ -126,6 +126,66 @@ macro_rules! test_ptx_warp {
     };
 }
 
+macro_rules! test_ptx_warp_with_input {
```
vosen: this is too complicated for what it does, either add input parameter to `test_ptx_warp` or use constants in the .ptx for inputs

vosen (changes requested):

ptx/src/test/spirv_run/mod.rs
AmosKito1: I extended `test_ptx_warp` with an input form and removed the separate `test_ptx_warp_with_input` macro. The regression still loads the delta values dynamically; the focused AMD test passes 4/4.

vosen: I probably did not make myself clear, but what I want is for input and output to the `test_ptx_warp!` to effectively a constant. That's how all the other tests are and it makes it easier to tell how a failing test is failing and what is expected of it. Currently the output is computed by `shfl_sync_dynamic_delta_expected` which is not what I want

AmosKito1: Sorry about the confusion, I understand what you meant now. I’ve updated the tests to use constants in the PTX fixtures and literal expected outputs in `test_ptx_warp!`. The extra input handling and calculated expected-output code have both been removed, leaving the existing test helper unchanged. All four focused AMD tests pass.

PR #678, opened 2026-09-12
Accept negative decimal floating-point PTX immediates

phantoslayer21:
PTX containing a negative decimal floating-point immediate is currently rejected. For example, both `.global .f64 negative = -1.5;` and `mov.f64 %fd, -1.5;` fail to parse.

Add a narrow immediate-parser alternative for a minus token followed by a decimal floating-point token. The existing unsigned float parser stays unchanged, including its use by `.version`. Regression coverage checks negative zero, signed exponents, integer backtracking, positive exact hexadecimal float/double constants, and a complete module containing global initializers and an instruction operand.

This implements the negative decimal case described by the [PTX constant-expression rules](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#constant-expressions); it does not add general constant-expression evaluation.

Validation on Windows with Rust 1.98.1, `x86_64-pc-windows-gnu`:
- `cargo test --locked -p ptx_parser`: 22 tests pass.
- The same new tests against unmodified production code at `9c8b43f2`: both new tests fail; the existing 20 tests pass.
- Rust formatting and `git diff --check` pass.

Opened as a draft for maintainer review. Full MSVC builds and GPU execution have not been run.

AI assistance: This contribution was developed with Codex and supervised GPT-5.6 Luna sub-agents for investigation, implementation, and review. The validation results above describe the tests actually run.


vosen (changes requested):
That's an ok idea, but it should be done in the lexer instead

phantoslayer21:
Thanks, I’ll move this into the lexer.

PR #679, opened 2026-09-12
Validate precompiler section and fatbin frame bounds

phantoslayer21:
The precompiler reads ELF section names from unchecked offsets and trusts each fatbin frame size before scheduling a load. Invalid section offsets can panic, names can extend beyond the declared string table, and a zero-sized frame can repeatedly schedule the same frame without advancing.

Validate ELF string-table and section ranges with checked conversions and arithmetic. Add a bounded frame iterator and use its validated ranges in the actual extraction loop before scheduling CUDA calls. Each accepted frame must contain its header and declared payload and advance the iterator. Valid concatenated frames, extended headers, empty payloads, and trailing padding remain supported.

Validation on Windows with Rust 1.98.1, `x86_64-pc-windows-gnu`:

- `cargo test --locked -p zluda_precompile`: all 13 tests pass.
- Six ELF tests cover valid extraction, invalid offsets/names, missing terminators, and overflow. Against original production code at `9c8b43f2`, four of those six tests fail.
- Seven frame tests cover concatenation, nonzero section offsets, extended headers, empty payloads, trailers, invalid ranges, zero/short headers, overflow, and truncation.
- Rust formatting and `git diff --check` pass.

The existing ELF class/endianness handling and PE section discovery remain unchanged. This validates outer frame boundaries, not every nested fatbin record. Full MSVC builds and end-to-end GPU precompilation have not been run.

AI assistance: This contribution was developed with Codex and supervised GPT-5.6 Luna sub-agents for investigation, implementation, and review. The validation results above describe the tests actually run.


vosen:
Does it cover any existing binaries? If it's purely to guard against theoretical problems then I'll close it, that app simply does not require that level of robustness

Issue #655, opened 2026-07-06
Improve FP64 accuracy


vosen:
Expected outcome:
All fp64 operations are bit-accurate as defined by PTX spec (https://docs.nvidia.com/cuda/parallel-thread-execution/index.html)

Comments:
* fp32 and lower is already accurate 
* fp64 is pretty much only used in HPC, so there's been no pressing need to improve it. With ZLUDA not supporting sever gpus this has been somewhat low priority, but since mi450x is shaping up to be supported by ZLUDA, it's getting more relevant
* We have a test suite here to test it: https://github.com/vosen/ptx_tests

DJHanceNL:
its really awesome you are still working on this. even though in spare time. Amazing skill. 

Interested what the real life use case for 16/19 decimal fp64 models/interference would be, maybe my local weather provider should switch, because they are always wrong

vosen:
By the 16/19 you mean fp16/bfloat16/tf32 machine learning models? fp64 is unrelated to machine learning. It's used in simulations: physics, chemistry, etc

DJHanceNL:
Yes, they have 16 or 19 decimal point numbers. The precision. 

Now I know. It's for science and chemistry extreme accuracy

PR #680, opened 2026-09-19
Fix rsqrt.approx.ftz.f64 lowering

somtri:
Related to #655.

Handle `rsqrt.approx.ftz.f64` separately from normal f64 rsqrt. The lowering clears the input's low word before `rsq`, clears the result's low word afterward, and uses PTX's canonical NaN value for NaN inputs.

The regression uses an f64 whose low word is nonzero. FTZ reduces it to 1.0, so the expected result is 1.0.

Verified locally:
- `test::spirv_run::rsqrt_ftz_f64_llvm`
- `test::spirv_run::rsqrt_ftz_f64_cuda`
- Rust 2021 formatting for the changed Rust file
- `git diff --check`

The available Radeon iGPU cannot run HIP workloads. HIP fails before ZLUDA starts, so AMD execution remains for CI or supported hardware.


vosen:
What is this code supposed to fix? As far as I can tell it both makes `rsqrt.approx.ftz.f64` slower and less accurate?

somtri:
This is for the special `rsqrt.approx.ftz.f64` opcode, not normal f64 rsqrt.

PTX defines this opcode as an approximation. It uses `a[63:32]` and clears `d[31:0]`, so its result is intentionally coarser than normal f64 rsqrt. The earlier lowering passed the full f64 value to `llvm.amdgcn.rsq.f64`, which did not implement this opcode.

The first revision also missed signed flushing of subnormal inputs. Commit `2754cc0` maps positive and negative f64 subnormals to signed zero before `rsq`, which produces `+Inf` and `-Inf`. The focused LLVM snapshots and CUDA-oracle tests cover normal input and both subnormal signs.

I do not have an AMD timing measurement. This adds classification work, so it may be slower than the prior lowering. The prior lowering was not semantically correct for `rsqrt.approx.ftz.f64`.

PTX specification: https://docs.nvidia.com/cuda/parallel-thread-execution/#floating-point-instructions-rsqrt-approx-ftz-f64

vosen:
closing for slop

